PR #3, opened 2024-03-29
FEAT minimum working version

Charlie-XIAO:
WIP

### DIFF
--- a/src-tauri/src/bundler.rs
+++ b/src-tauri/src/bundler.rs
@@ -1,0 +1,299 @@
+use std::{collections::HashMap, path::{Component, Path, PathBuf}};
+
+use anyhow::{bail, Error};
+use path_clean::PathClean;
+use serde::Serialize;
+use swc_atoms::Atom;
+use swc_bundler::{Bundler, Hook, Load, ModuleData, ModuleRecord, Resolve};
+use swc_common::{
+    comments::SingleThreadedComments,
+    errors::{ColorConfig, Handler},
+    sync::Lrc,
+    FileName, FilePathMapping, Globals, Mark, SourceMap, Span, GLOBALS,
+};
+use swc_ecma_ast::KeyValueProp;
+use swc_ecma_codegen::{text_writer::{JsWriter, WriteJs}, Emitter};
+use swc_ecma_loader::resolve::Resolution;
+use swc_ecma_parser::{parse_file_as_module, EsConfig, Syntax};
+use swc_ecma_transforms_react::jsx;
+use swc_ecma_visit::FoldWith;
+
+#[cfg(windows)]
+use normpath::BasePath;
+
+// The file extensions that are recognized by the bundler
+static EXTENSIONS: &[&str] = &["js", "jsx", "ts", "tsx"];
+
+pub(crate) enum BundlerOutput {
+    Code(String),
+    Error(String),
+}
+
+pub(crate) fn bundle(
+    target: &PathBuf,
+    dependency_map: Option<&HashMap<String, String>>,
+) -> BundlerOutput {
+    match bundle_internal(target, dependency_map) {
+        Ok(code) => BundlerOutput::Code(code),
+        Err(err) => BundlerOutput::Error(format!("{:#?}", err)),
+    }
+}
+
+// Treat the target file as an entry point and bundle it into a single module that can
+// be recognized by the <script> tag in a browser; `dependency_map` is the mapping that
+// should be read from the configuration file
+fn bundle_internal(
+    target: &PathBuf,
+    dependency_map: Option<&HashMap<String, String>>,
+) -> Result<String, Error> {
+    let globals = Globals::default();
+    let cm = Lrc::new(SourceMap::new(FilePathMapping::empty()));
+
+    // Get the external modules that should not be resolved and bundled, but rather be
+    // left as `import` statements as is
+    let external_modules = match dependency_map {
+        Some(map) => map.keys().map(|k| Atom::from(k.clone())).collect(),
+        None => vec![],
+    };
+
+    let mut bundler = Bundler::new(
+        &globals,
+        cm.clone(),
+        PathLoader(cm.clone()),
+        PathResolver,
+        swc_bundler::Config { external_modules, ..Default::default() },
+        Box::new(NoopHook),
+    );
+
+    // Bundle the target file into bundles, each containing the merged module
+    let mut entries = HashMap::new();
+    entries
+        .insert(target.to_string_lossy().to_string(), FileName::Real(target.clone()));
+    let mut bundles = bundler.bundle(entries)?;
+
+    // Since we provide a single entry point, there should be only one generated
+    // bundle (as long as there are no dynamic/conditional imports)
+    if bundles.len() != 1 {
+        bail!("Expected a single bundle; try to remove dynamic/conditional imports");
+    }
+    let module = bundles.pop().unwrap().module;
+
+    let code = GLOBALS.set(&globals, || {
+        // Notes on the JSX transform
+        // ==========================
+        //
+        // There are two types of JSX transforms ("classic" and "automatic"), see
+        // https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html
+        //
+        // The "automatic" transform automatically imports from "react/jsx-runtime", but
+        // this module is not available when runnning the bundled code in the browser,
+        // so we have to use the "classic" transform instead. The "classic" transform
+        // requires `React` to be in scope, which we can require users to bring into
+        // scope by assigning `const React = window.__DESKULPT__.React`.
+        //
+        // Note, however, that this puts constraints on how we can minify the bundled
+        // code, e.g., we cannot mangle the `React` identifier, we cannot remove `React`
+        // even if it is unused, etc.
+
+        let mut jsx_transform = jsx::<SingleThreadedComments>(
+            cm.clone(),
+            None,
+            Default::default(), // options, where runtime defaults to "classic"
+            Mark::new(),        // top level mark
+            Mark::new(),        // unresolved mark
+        );
+
+        // TODO: maybe need to chain more transforms, e.g., TypeScript transform
+        let module = module.fold_with(&mut jsx_transform);
+
+        // Emit the bundled module as string into a buffer
+        let mut buf = vec![];
+        {
+            let wr = JsWriter::new(cm.clone(), "\n", &mut buf, None);
+            let mut emitter = Emitter {
+                cfg: swc_ecma_codegen::Config::default().with_minify(true),
+                cm: cm.clone(),
+                comments: None,
+                wr: Box::new(wr) as Box<dyn WriteJs>,
+            };
+            emitter.emit_module(&module).unwrap();
+        }
+        String::from_utf8_lossy(&buf).to_string()
+    });
+
+    Ok(code)
+}
+
+struct PathLoader(Lrc<SourceMap>);
+
+impl Load for PathLoader {
+    fn load(&self, file: &FileName) -> Result<ModuleData, Error> {
+        let fm = match file {
+            FileName::Real(path) => self.0.load_file(path)?,
+            _ => unreachable!(),
+        };
+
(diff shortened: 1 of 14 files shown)


src-tauri/src/main.rs
```diff
@@ -1,24 +1,33 @@
...
+/// Main entry point of Deskulpt.
 fn main() {
-    tauri::Builder::default()
-        .setup(|app| {
-            deskulpt_setup(app).context("\nFailed to setup Deskulpt")?;
-            Ok(())
-        })
+    // Get the widget base directory in advance; it seems that `.setup` may not finish
+    // before the frontend is loaded, causing errors like accessing unmanaged state
+    let context = generate_context!();
```
ROMEEZHOU, line 16: error when running `npm run tauri dev`:
```
error: symbol `_EMBED_INFO_PLIST` is already defined
  --> src/main.rs:31:14
   |
31 |         .run(generate_context!())
   |              ^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `$crate::embed_info_plist_bytes` which comes from the expansion of the macro `generate_context` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `deskulpt` (bin "deskulpt") due to 1 previous error
```

Charlie-XIAO: This is caused by using the `generate_context` macro twice, which seems not to be designed for multiple calls. Anyways, reusing `context` in the second call should solve the issue.

P.S. Seems that error does not occur on Windows.

Xinyu-Li-123:
Under which condition should we merge this pull request? Are Deskulpt supposed to successfully render all test widgets in [this repo ](https://github.com/CSCI-SHU-410-SE-Project/test-widgets) in this pull request?

Charlie-XIAO:
> Under which condition should we merge this pull request?

AFAIC,

1. Test on the three platforms, see whether there are weird behaviors that you do not expect.
2. Review whether my code organization is reasonable (since we want to avoid large-scale refactoring to our best effort).
3. Review whether there are places where I write confusing code.
4. Discuss (if any) more things we should include in this PR. But note that we should keep the PR small, also due to limited time.

If all the above points are addressed the PR would be in general good to go.

> Are Deskulpt supposed to successfully render some of the failed test widgets in [this repo](https://github.com/CSCI-SHU-410-SE-Project/test-widgets) in this pull request?

No, as described in the README of each widget they are meant to fail, and Deskulpt should render proper error display for them.

> Also, is the todo list supposed to be for the project in general or for this pull request alone?

The TODO list is not for the PR, which contains things that we may want to do but will not (if people agree) do in this PR. After this PR is merged, we will get a better idea of the next-steps, and maybe we can open a meta-issue specifically for that.

Charlie-XIAO:
By the way, the goal is that core developers at least have a basic understanding of *where to find things*, i.e., if I want to modify behavior A, which part of code I may need to change. This painful process of *understanding others' code* need to come first, otherwise follow-up PRs will be hard.

Xinyu-Li-123:
# Security Issue: Widget can import from any files on host file system
When messing with the test widgets, I find that **a widget can import from arbitrary files, even those outside of `$APPDATA`**. Below is an example of a clock widget that can access
- utils of other widget
- utils of Deskulpt
- utils of another app (not Deskulpt)
```typescript
// File: $AppData/widgets/ok-counter/src/App.jsx

const React = window.__DESKULPT__.defaultDeps.React;

// $AppData/widgets/ok-clock/src/utils/utils.js
import okCounterWhisper from "./utils/utils";
// $AppData/widgets/ok-counter/src/utils/utils.js
import { okClockShout } from "../../ok-clock/src/utils/utils";
// $AppData/secret.js
import { secretFunc as deskulptSecretFunc } from "../../../deskulptSecret"
// C:\Users\eiger\AppData\Roaming\
import { secretFunc as myOtherAppSecretFunc } from "../../../../MyOtherApp/myOtherAppSecret"


function Counter() {
  const [count, setCount] = React.useState(0);
  return (
    <div>
      <p>{okCounterWhisper("My Counter")}</p>
      <p>{okClockShout("My Counter")}</p>
      <p>{deskulptSecretFunc()}</p>
      <p>{myOtherAppSecretFunc()}</p>
      <p>Count: {count}</p>
      <button onClick={() => setCount(count + 1)}>Increment</button>
    </div>
  );
}

export default {
  render: () => <Counter />,
};
```

And this is the rendered result
image

I think this should be solved within this PR, or at least we should figure out how to solve it using rust and tauri's functionality.

Charlie-XIAO:
So the problem is that with relative imports we can go to parents and eventually outside the widget directory. We may be able to raise in the path resolver if the resolved path is not a child of the widget directory.

I agree that we can solve it in this PR because it does not seem to be very hard.

Xinyu-Li-123:
# "Render All" button won't remove deleted / ignored widgets

If a widget is rendered, and later deleted or ignored, it won't be cleared from the canvas. Below is what I have tested:
- Delete a rendered widget:
  - add widget to `$APPDATA/widgets`; 
  - click `rescan` and `render all`; 
  - remove the widget folder; 
  - click `rescan` and `render all`;
- Ignore a rendered widget: 
  - add widget to `$APPDATA/widgets`; 
  - click `rescan` and `render all`; 
  - change the `ignore` option in `deskulpt.conf.json` from `false` to `true`
  - click `rescan` and `render all`;

Charlie-XIAO:
> So the problem is that with relative imports we can go to parents and eventually outside the widget directory. We may be able to raise in the path resolver if the resolved path is not a child of the widget directory.
> 
> I agree that we can solve it in this PR because it does not seem to be very hard.

I added a few bad widgets in the widget repo, for which the bundler rejects their import statements. [This one](https://github.com/CSCI-SHU-410-SE-Project/test-widgets/tree/main/err-bundler-import-beyond-root) should represent your case @Xinyu-Li-123, and it would be nice if you can confirm on your side as well. [`5e7a3d4`](https://github.com/CSCI-SHU-410-SE-Project/Deskulpt/pull/3/commits/5e7a3d4f2f226836169536ab3b49f2a0df82d327) is the commit that made the fix.

Xinyu-Li-123:
> > So the problem is that with relative imports we can go to parents and eventually outside the widget directory. We may be able to raise in the path resolver if the resolved path is not a child of the widget directory.
> > I agree that we can solve it in this PR because it does not seem to be very hard.
> 
> I added a few bad widgets in the widget repo, for which the bundler rejects their import statements. [This one](https://github.com/CSCI-SHU-410-SE-Project/test-widgets/tree/main/err-bundler-import-beyond-root) should represent your case @Xinyu-Li-123, and it would be nice if you can confirm on your side as well. [`5e7a3d4`](https://github.com/CSCI-SHU-410-SE-Project/Deskulpt/pull/3/commits/5e7a3d4f2f226836169536ab3b49f2a0df82d327) is the commit that made the fix.
> 


I think your fix may have a bug. Now I can't import files within the src directory using relative path. The clock app below won't be bundled correctly

```typescript
// File: $APPDATA/widgets/ok-clock/src/App.jsx

const React = window.__DESKULPT__.defaultDeps.React;

// $APPDATA/widgets/ok-clock/src/utils/utils.js
import { okClockShout } from "./utils/utils";

const options = {
  year: "numeric",
  month: "short",
  day: "numeric",
  hour: "numeric",
  minute: "numeric",
};

function Clock() {
  const [time, setTime] = React.useState(new Date().toLocaleString("en", options));

  function updateTime() {
    setTime(new Date().toLocaleTimeString("en", options));
  }
  setInterval(updateTime, 10000);

  return (
    <div>
      <p>{okClockShout("My Clock")}</p>
      <p>{time}</p>
    </div>
  );
}

export default {
  render: () => <Clock />,
};

```

The error log in the canvas is as follows
```
[Backend] Widget (id=ok-clock) fails to be bundled
Error {
    context: "Failed to bundle widget (id=ok-clock)",
    source: Error {
        context: "load_transformed failed",
        source: Error {
            context: "failed to analyze module",
            source: Error {
                context: "failed to resolve ./utils/utils from \\\\?\\C:\\Users\\eiger\\AppData\\Roaming\\com.tauri.deskulpt\\widgets\\ok-clock\\src\\App.jsx",
                source: "Relative imports should not go beyond the root \"C:\\\\Users\\\\eiger\\\\AppData\\\\Roaming\\\\com.tauri.deskulpt\\\\widgets\\\\ok-clock\"",
            },
        },
    },
}
```

Charlie-XIAO:
It is the problem with the Windows `\\?\` long-path prefix. [`d80d00f`](https://github.com/CSCI-SHU-410-SE-Project/Deskulpt/pull/3/commits/d80d00fe0eef09629ec6887ce738f42b774d70e9) fixes via canonicalization, but I'm not sure if this will cause issues when there are symlinks. But those should be rare cases and we may not need to take care of it at this moment. I added a comment in the code as a reminder.

Xinyu-Li-123:
> It is the problem with the Windows `\\?\` long-path prefix. [`d80d00f`](https://github.com/CSCI-SHU-410-SE-Project/Deskulpt/pull/3/commits/d80d00fe0eef09629ec6887ce738f42b774d70e9) fixes via canonicalization, but I'm not sure if this will cause issues when there are symlinks. But those should be rare cases and we may not need to take care of it at this moment. I added a comment in the code as a reminder.

And it works on my machine as well. Also, I agree that we should worried about symlink later, since we may change the rust code later.

src-tauri/src/bundler.rs
```diff
@@ -0,0 +1,331 @@
...
+
+impl Hook for NoopHook {
+    fn get_import_meta_props(
+        &self,
+        _: Span,
+        _: &ModuleRecord,
+    ) -> Result<Vec<KeyValueProp>, Error> {
+        unimplemented!();
+    }
+}
```
ROMEEZHOU, line 331: Why we have to write a no operation hook here. Any existing example of using hook in bundler?

src-tauri/src/bundler.rs
```diff
@@ -0,0 +1,331 @@
...
+        swc_bundler::Config { external_modules, ..Default::default() },
+        Box::new(NoopHook),
+    );
+
+    // Bundle the target file into bundles, each containing the merged module
+    let mut entries = HashMap::new();
+    entries.insert(
+        target.to_string_lossy().to_string(),
+        FileName::Real(target.to_path_buf()),
+    );
```
ROMEEZHOU, line 76: By the way, what specifically does the hashmap do here

src-tauri/src/bundler.rs
Charlie-XIAO: Example: https://github.com/swc-project/swc/blob/main/crates/swc_bundler/examples/bundle.rs. Though there is completely no documentation for this, there are indeed some discussions in https://github.com/swc-project/swc/pull/1195. We may want to keep this in mind, but I think `import.meta` is a rare use-case for widgets. Since it is explicitly marked as `unimplemented` I believe we can make a fix until it is really hit by someone.

src-tauri/src/bundler.rs
Charlie-XIAO: Comment updated.

Charlie-XIAO:
@ROMEEZHOU @Xinyu-Li-123 If you are okay with the current state of the PR, I think we can merge so that further work won't be blocked for too long :)

Xinyu-Li-123 (approved):
I've reviewed the changes, and everything looks good. This PR implements the core functionality of "bundling widget code at runtime of Deskulpt".

Charlie-XIAO:
#6 tracks the unresolved https://github.com/CSCI-SHU-410-SE-Project/Deskulpt/pull/3#issuecomment-2036444329, I'm going to squash and merge. Thanks for the reviews!

rz2606:
npm run tauri build -- --debug  # production build
./src-tauri/target/debug/deskulpt.exe  # may differ  

run deskulpt.exe  page is empty

Charlie-XIAO:
Can you show a screenshot and of the two windows, and your console output?

rz2606:
I want to debug in  rustrover, how to configure settings?

Charlie-XIAO:
I'm not sure how `rustrover` is used, but to understand your issue, I need to see what the two app windows look like on your machine, and the output in the terminal in which you run `./src-tauri/target/debug/deskulpt.exe`.

PR #5, opened 2024-04-01
MAINT `cargo-clippy` is not included in `pre-commit` and ignore `eslint` warning for ignored files

Charlie-XIAO:
As described in the title.

/cc @CSCI-SHU-410-SE-Project/core-dev (testing GitHub teams and requesting review)


Xinyu-Li-123 (approved):
Valid point.

PR #7, opened 2024-04-05
FIX bundler removing `React` when not explicitly used in code

Charlie-XIAO:
If one creates a simple widget like this:

```jsx
const React = window.__DESKULPT__.defaultDeps.React;

export default {
  render: () => <h1>Welcome!</h1>,
};
```

The bundler removes the `React` assignment because it performs tree shaking by default [[ref](https://rustdoc.swc.rs/src/swc_bundler/bundler/optimize.rs.html#22-33)] and `React` is not explicitly used in the code. However, we require to bring `React` into scope since the (classical) JSX transform  relies on it to transpile.

The workaround is to disabled tree-shaking in the bundler and fold with tree-shaking optimization later by ourselves. This way we can specify `React` as `top_retain` to avoid it from being removed by unused.

Later I will make sure to test this case when the test suite is set up.

/cc @CSCI-SHU-410-SE-Project/core-dev for review.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -703,6 +703,7 @@ dependencies = [
  "swc_ecma_loader",
  "swc_ecma_minifier",
  "swc_ecma_parser",
+ "swc_ecma_transforms_optimization",
  "swc_ecma_transforms_react",
  "swc_ecma_visit",
  "tauri",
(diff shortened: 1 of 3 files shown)


ROMEEZHOU (approved):
works well on my machine

PR #8, opened 2024-04-06
MAINT workaround pre-commit `/dev/tty` not found and fix escape error on Windows CMD

Charlie-XIAO:
This PR does the following:

- Redirect only when file descriptor 2 (stderr) is connected to the terminal. This workaround is suggested in https://github.com/typicode/husky/issues/968#issuecomment-1184296264. I've tested with the VS Code "commit" button; it raises previously and works now.
- Better escape mechanism in `lint-staged`. This part refers to [the configuration of `next.js`](https://github.com/vercel/next.js/blob/canary/lint-staged.config.js).

@Xinyu-Li-123 Can you please confirm if these solve your issue?

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -5,5 +5,8 @@
 # otherwise the fancy output of lint-staged will be disabled; see the husky issue
 # https://github.com/typicode/husky/issues/968
 
-exec 1> /dev/tty
+if [ -t 2 ]; then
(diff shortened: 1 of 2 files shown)


Xinyu-Li-123 (approved):
I have test `git commit` on vscode git gui, windows command prompt, and git bash on Windows 10, and the modified pre-commit hooks works smoothly.

PR #10, opened 2024-04-06
TST unit tests for bundler

ROMEEZHOU:
I have written some unit tests for bundler, which cover the following cases:
- bundling a simple file that doesn't explicitly use `React`
- bundling a simple file that explicitly uses `React`
- bundling a file that imports from a `.js` file using relative path with extension, e.g. `import { getMessage } from "./utils.js";`,
- bundling a file that imports from a `.js` file using relative path without extension
- bundling a file that imports from a directory using relative path
- bundling a file that imports from a `.jsx` file using relative path without extension, in this case, both the file being imported and the main file has `const React = window.__DESKULPT__.defaultDeps.React;`
- bundling a file that imports from a `.jsx` file, the file being imported doesn't define `React`
- bundling a file that imports from a `.jsx` file ,the main file doesn't define `React`
- bundling a file that imports using an absolute path (which should raise errors)
- bundling a file that imports from a file that is beyond the root (which should raise errors)

You can `cd` to `src-tauri` and run the test by 
```
cargo test
```
 You may also check the coverage visualization by running 
```
cargo llvm-cov --open 
```

@CSCI-SHU-410-SE-Project/core-dev  Please let me know if there are any problems :)


src-tauri/src/bundler.rs
```diff
@@ -337,3 +337,84 @@ impl Hook for NoopHook {
...
+
+    #[parameterized(case = {
+        "no_react",  // Not explicitly using React
+        "with_react_hook",  // Using React hook
+        "import_relative",  // Relative import with extension
+        "import_relative_no_ext",
+        "import_relative_directory",
+        "import_jsx",
+        "import_no_inner_React",
+        "import_no_outer_React"
```
Charlie-XIAO, line 369: Can you add short descriptions for these as well, similar to the three test cases above?

By the way, I would suggest renaming (the directory names also need to be updated correspondingly):

- `import_jsx` -> `import_relative_no_ext_jsx`
- `import_no_inner_React` -> `import_no_inner_react`
- `import_no_outer_React` -> `impmort_no_outer_react`

src-tauri/src/bundler.rs
```diff
@@ -337,3 +337,84 @@ impl Hook for NoopHook {
...
+            format!("failed to resolve /usr/bin/script.js from {root_path}/index.jsx"),
+            "Absolute imports are not supported; use relative imports instead"
+                .to_string(),
+        ];
+        assert_err_eq(error, expected);
+    }
+
+    #[test]
+    fn test_bundle_beyond_root_error() {
+        // Test that we do not allow absolute path import
```
Charlie-XIAO, line 405:
```suggestion
    fn test_bundle_relative_import_beyond_root_error() {
        // Test that we do not allow relative path imports to go beyond the root
```

src-tauri/src/bundler.rs
```diff
@@ -337,3 +337,84 @@ impl Hook for NoopHook {
...
+        let root_path = "tests/fixtures/bundler/import_beyond_root/input";
+        let root = PathBuf::from(root_path);
+        let error = bundle(&root, root.join("index.jsx").as_path(), None)
+            .expect_err("Expected an error");
+        let abs_root = root.canonicalize().expect("Failed to get absolute path");
```
Charlie-XIAO, line 410:
```suggestion
        let abs_root = root.canonicalize().unwrap();
```

I would suggest using `unwrap` here since it is not directly related to the function we test.

src-tauri/tests/fixtures/bundler/import_absolute/input/index.jsx
```diff
@@ -0,0 +1,15 @@
...
+  return (
+    <div>{getMessage()}</div>
+  );
+}
+
+const App = {
+  render: () => <Message />,
+};
+
+export default App;
```
Charlie-XIAO, line 15: We don't really need all these subsequent lines, since we are only testing error for absolute import. I would prefer keeping it minimal.

src-tauri/tests/fixtures/bundler/import_beyond_root/input/index.jsx
```diff
@@ -0,0 +1,15 @@
...
+  return (
+    <div>{getMessage()}</div>
+  );
+}
+
+const App = {
+  render: () => <Message />,
+};
+
+export default App;
```
Charlie-XIAO, line 15: Same here, no need for these lines.

src-tauri/tests/fixtures/bundler/import_beyond_root/input/index.jsx
```diff
@@ -0,0 +1,15 @@
+import { getMessage } from "../../../import_jsx/input/utils";
```
Charlie-XIAO, line 1: I'm interested in what will happen if this import path does not exist. When I wrote the code I remember that I used canonicalization somewhere, which may cause this to fail not with the expected failure?

src-tauri/tests/fixtures/bundler/import_beyond_root/output.js
```diff
@@ -0,0 +1 @@
+function getMessage(){return"This function is successfully imported.";}const React=window.__DESKULPT__.defaultDeps.React;function Message(){return React.createElement("div",null,getMessage());}const App={render:()=>React.createElement(Message,null)};export{App as default};
```
Charlie-XIAO, line 1: `import_beyond_root` seems to be a failing case, I think we do not need this file?

src-tauri/tests/fixtures/bundler/import_jsx/input/index.jsx
```diff
@@ -0,0 +1,15 @@
+import getMessage from "./utils";
+
+const React = window.__DESKULPT__.defaultDeps.React;
+
+function Message() {
+  return (
+    <div>{getMessage()}</div>
+  );
```
Charlie-XIAO, line 8:
```suggestion
  return <div><MessageInner /></div>;
```

NIT code style.

src-tauri/tests/fixtures/bundler/import_jsx/input/utils.jsx
```diff
@@ -0,0 +1,10 @@
+const React = window.__DESKULPT__.defaultDeps.React;
+
+export default function getMessage(){
+  return(
+    <div>
+      <p>This React component is successfully imported.</p>
+    </div>
+  );
+}
```
Charlie-XIAO, line 9:
```suggestion
export default function MessageInner() {
  return <div>This React component is successfully imported.</div>;
}
```

NIT code style, and please follow the convention that React functional components are capitalized.

src-tauri/tests/fixtures/bundler/import_jsx/input/index.jsx
```diff
@@ -0,0 +1,15 @@
+import getMessage from "./utils";
```
Charlie-XIAO, line 1:
```suggestion
import getMessage from "./MessageInner";
```

We commonly name the file of a React component exactly by the name of that component,

src-tauri/tests/fixtures/bundler/import_no_inner_React/input/index.jsx
```diff
@@ -0,0 +1,15 @@
+import getMessage from "./utils";
```
Charlie-XIAO, line 1: Same for this test case.

src-tauri/tests/fixtures/bundler/import_no_outer_React/input/index.jsx
```diff
@@ -0,0 +1,13 @@
+import getMessage from "./utils";
```
Charlie-XIAO, line 1: Same for this test case.

src-tauri/tests/fixtures/bundler/import_relative/input/index.jsx
```diff
@@ -0,0 +1,15 @@
+import { getMessage } from "./utils.js";
+
+const React = window.__DESKULPT__.defaultDeps.React;
+
+function Message() {
+  return (
+    <div>{getMessage()}</div>
+  );
```
Charlie-XIAO, line 8:
```suggestion
  return <div>{getMessage()}</div>;
```

NIT code style.

src-tauri/tests/fixtures/bundler/import_relative_directory/input/index.jsx
```diff
@@ -0,0 +1,15 @@
+import { getMessage } from "./utils";
+
+const React = window.__DESKULPT__.defaultDeps.React;
+
+function Message() {
+  return (
+    <div>{getMessage()}</div>
+  );
```
Charlie-XIAO, line 8:
```suggestion
  return <div>{getMessage()}</div>;
```

NIT code style.

src-tauri/tests/fixtures/bundler/import_relative_no_ext/input/index.jsx
```diff
@@ -0,0 +1,15 @@
+import { getMessage } from "./utils";
+
+const React = window.__DESKULPT__.defaultDeps.React;
+
+function Message() {
+  return (
+    <div>{getMessage()}</div>
+  );
```
Charlie-XIAO, line 8:
```suggestion
  return <div>{getMessage()}</div>;
```

NIT code style.

Charlie-XIAO (commented):
Thanks for the PR @ROMEEZHOU, this is impressive work!

Please see the review comments for my suggestions. Once they are resolved we can check if we want more test cases in this PR or postpone to further PRs to refine the test suite.

src-tauri/src/bundler.rs
ROMEEZHOU: Please check the latest commit for revised version :)

Charlie-XIAO:
LGTM, thanks @ROMEEZHOU! We shall take care of code coverage and more error cases in follow-up PRs.

Charlie-XIAO (approved):

PR #12, opened 2024-04-07
FEAT Deskulpt Widget API

Xinyu-Li-123:
This PR defines the basic architecture of Deskulpt widget apis. Please note that the PR has maken some **backward-incompatible** changes on how a widget is defined in Deskulpt. 

## Widget Definition: Migrating from JS Script to React Class Component

Currently, a widget source code is a javascript file that exposes a `render() -> React.JSX.Element` method. In this PR, a widget source code should define a React class component instead. There are two motivations of this migration:
1. to provide widget apis with access to widget-specific data, such as widget id.
2. to provide widgets with access to React component lifecycles (e.g. do something on component mount)

### 1. Access to Widget-specific Data

Widget apis are tauri commands exposed to widgets. Conceptually, to invoke a widget api, the api's implementation must have access to the widget id. For example, to read a file from the widget storage, the `read_file` api` need to know the widget's id to craft the absolute path to the file. As a result, the function signature of `read_file` should be something like
```rust
#[tauri::command]
fn read_file( widget_id: String, path: String ) -> String;
// read file from $APPDATA/widgets/<widget_id>/storage/<path>
```
However, using our current definition of widget, i.e. a jsx script that exports a `render() -> React.JSX.Element` method, there is no clear way for a widget to explicitly pass widget id to widget api. To sum up, we need a way to bind functions (widget api `readd_file`) with data (`widget_id`).

A natural solution is class, where `widget_id` can be defined as a private property of the `Widget` class, and passed to `read_file` method internally. We would have a base `Widget` class like this
```typescript
class Widget: 
  // "#" means private variable
  #widget_id: string = "";
  
  constructor(widget_id: string) {
    this.#widget_id = widget_id;
  }

  // invocation of `read_file` api is wrapped in a method so that it has access to `widget_id`
  async read_file(path: string) {
    const result = await invoke("read_file", { widget_id: this.#widget_id, path: path });
    ...
  }

  // to be overridden by subclass widgets
  render() {
    throw new Error("Not implemented")
  }
```
A widget developer may extend this base widget class to define their own widget.

### 2. Access to React Component Lifecycle

However, class alone is not enough if we want the widget to have access to React component's lifecycle, such as `componentDidMount` [^1]. This is a necessary feature for even simple widgets. For example, a todolist widget need to read a `todo.txt` file that contains all todo items right after the todolist component is mounted.

This motivates us to define a widget as a subclass of `React.Component`. The base class `Widget` would be something like this
```typescript
interface WidgetProps {
  id: string;
}

class Widget extends React.Component<WidgetProps> {
  widget_api = {
    fs: {
      read_file: async (path: string) => {
        try {
          const result = await invoke("read_file", {
            // widget id is passed in as a prop, and accessed via `this.props.id`
            widgetId: this.props.id,
            path: path,
          });
          return result;
        } catch (error) {
          return error;
        }
      }}};

  // to be overridden by subclass widgets
  render(): React.JSX.Element {
    return <h1>Widget {this.props.id}</h1>;
  }
}
```

## Widget API

In this PR, I create two sets of widget apis: `dummy` and `fs` as a demo.
```
plugin:
   name: "widget_api.dummy"
   commands:
      get_dummy_info(widget_id: String) -> String

plugin:
    name: "widget_api.fs"
    commands:
      // read files in the storage of a widget
      read_file(app_handle: AppHandle, widget_id: String, path: String) -> Result<String, String>
```
As an example, to invoke `read_file` in Javascript, we can do this
```javascript
invoke("plugin:widdget_api.fs|read_file", {widget_id: "ok-poem-display", path: "poem.txt"})
```
This will read the content in the file `$APPDATA/widgets/ok-poem-display/storage/poem.txt`.

## How to Test this PR

Please first checkout the PR branch locally. For testing widgets, [this repo](https://github.com/CSCI-SHU-410-SE-Project/test-class-widgets) contains some simple widgets defined as React class components. 

[^1]: Although it's generally recommended to use functional component instead of class component in React, I think class component just naturally fits our usecase. The downside here is that we can't use hooks like `useEffect`, but most hooks have their counterparts in class component lifecycle. For example, we can replace `useEffect(handler, [])` with `componentDidMount`.


Charlie-XIAO:
Why do we have to use React components? What's the problem of requiring the `render()` to return a component taking `props: { id: string }` instead of no `props`?

Xinyu-Li-123:
> Why do we have to use React components? What's the problem of requiring the `render()` to return a component taking `props: { id: string }` instead of no `props`? Is it because we have to force a widget not to use other widgets' ID?
> 
> If that's the reason, can we just bind the widget APIs to `props` and let widgets access APIs via `props`, for instance,
> 
> ```js
> function Time(props: { api: DeskulptWidgetAPI }) {
>   const { api } = props;
> 
>   return <h1>{api.getTime()}</h1>;
> }
> ```

In our current implementation, I can't find a way for a widget to pass widget id as an argument to invocation of tauri command, at least, no maintainable way. It may help with the understanding if you try to write a widget that displays its own widget id on the canvas,

Charlie-XIAO:
What about:

```js
function shout(text) {
  console.log(this.id, text.toUpperCase());
}

const generalAPI = {
  shout,
}

const widgetAPI = Object.fromEntries(
  Object.keys(generalAPI).map(k => [k, (...args) => generalAPI[k](widgetId, ...args)])
);

widgetDOMRoot.react.render(
  <WidgetContainer id={widgetId}>
    <Widget api={widgetAPI} />
  </WidgetContainer>
);
```

Then a widget can access the `shout` API via `props.api.shout`?

Xinyu-Li-123:
That's interesting. I think you are on the right track. To associate data with functions, we can use either oop (attribute and method) or functional programming (partial application). Let me give it a try later today.

PR #13, opened 2024-04-07
FEAT Backend of widget fs api with rust unit test

Xinyu-Li-123:
This PR provides the architectural design of backend widget api. It contains rust implementation and rust unit tests of widget apis for file system access. Since this is a demo, we only focus on a preliminary set of fs api in this PR.

## What this PR is not about

This PR _doesn't provide any js wrapper for the widget apis_. Due to the discussion in #12 on the choice between functional or class React component, I think it's better to separate the backend widget api and the frontend widget api wrapper into two PRs. Both PR will be architectural ones that define how data flows in Deskulpt, instead of incremental ones that extend existing functionalities. And our choice between functional or class component in the frontend won't influence the backend widget apis implementation in the backend.

## Widget API

In this PR, I create two sets of widget apis: `dummy` and `fs` as a demo.
```
plugin:
   name: "widget_api.dummy"
   commands:
      get_dummy_info(widget_id: String) -> String

plugin:
    name: "widget_api.fs"
    commands:
      // read files in the storage of a widget
      read_file(app_handle: AppHandle, widget_id: String, path: String) -> Result<String, String>
      write_file(app_handle: AppHandle, widget_id: String, path: String, content: String) -> Result<(), String>
      append_file(app_handle: AppHandle, widget_id: String, path: String, content: String) -> Result<(), String>
      remove_file(app_handle: AppHandle, widget_id: String, path: String) -> Result<(), String>
```
As an example, to invoke `read_file` in Javascript, we can do this
```javascript
invoke("plugin:widdget_api.fs|read_file", {widget_id: "ok-poem-display", path: "poem.txt"})
```
This will read the content in the file `$APPDATA/widgets/ok-poem-display/storage/poem.txt`.

## Unit Test of Widget API

Unit tests of widget api is based on Rust's native unit test functionality. The test is restricted to backend api functionality, and doesn't test the interaction between frontend and backend (I think [Selenium](https://www.selenium.dev/) could do the job, but that deserves another PR). It simulates the tauri app using the [`mock_app`](https://docs.rs/tauri/latest/tauri/test/fn.mock_app.html) function provided by `tauri::test` module. To use this function, we need to enable the `test` feature of tauri.

Unit tests for fs api in specific uses the `tempfile` crate to tests in a temporary directory, which will be automatically created on test starts and deleted on test ends.


Xinyu-Li-123:
@Charlie-XIAO For modularity, I think this PR should only be responsible for designing the backend arch of widget api and related unit tests. I will work on the frontend wrapper of the backend widget api in a separate PR. I have rewritten the description of the PR, could you please check it out?

Charlie-XIAO:
Thanks for the update. I will take a look later today.

src-tauri/Cargo.toml
```diff
@@ -11,10 +11,11 @@ tauri-build = { version = "1.5.1", features = ["config-json5"] }
 [dev-dependencies]
 pretty_assertions = "1.4.0"
 parameterized = "2.0.0"
+tempfile = "3.2.0"
```
Charlie-XIAO, line 14: Latest version is `3.10.1` I believe. Is there a particular reason we pin a lower version?

src-tauri/src/main.rs
```diff
@@ -27,7 +28,11 @@ fn main() {
             commands::bundle_widget,
             commands::open_widget_base,
             commands::refresh_widget_collection,
+            // widget_api::dummy::get_dummy_info,
```
Charlie-XIAO, line 31: This should be removed.

src-tauri/src/main.rs
```diff
@@ -27,7 +28,11 @@ fn main() {
...
         ])
+        // initialize widget apis
```
Charlie-XIAO, line 33: We would prefer capitalized comments, even for inline comments that are not meant to be full sentences. Same for the other few places.

src-tauri/src/widget_api/README.md
```diff
@@ -0,0 +1,26 @@
+# Implementation of Widget API
```
Charlie-XIAO, line 1: Thanks for the development notes, but I think these can be somewhere else. Maybe we can move it to the wiki or open a meta-issue for tracking the development of widget API and move it there. WDYT?

src-tauri/src/widget_api/dummy/apis.rs
```diff
@@ -0,0 +1,11 @@
+use tauri::command;
```
Charlie-XIAO, line 1: What's the purpose of keeping a dummy demo which would definitely be removed in the future, given that we already have `widget_api::fs` as demo?

src-tauri/src/widget_api/fs/apis.rs
```diff
@@ -0,0 +1,154 @@
...
+use tauri::{command, AppHandle, Manager, Runtime};
+
+fn get_file_path<R: Runtime>(
+    app_handle: &AppHandle<R>,
+    widget_id: &str,
+    path: &str,
+) -> PathBuf {
+    let widget_base = &app_handle.state::<WidgetBaseDirectoryState>().0;
+    // if storage directory does not exist, create it
+    let widget_dir = widget_base.join(widget_id).join("storage");
```
Charlie-XIAO, line 13: I wonder why we force things to be in the `storage` directory. Isn't `widget_base.join(widget_id)` already enough?

src-tauri/src/widget_api/fs/apis.rs
```diff
@@ -0,0 +1,154 @@
...
+    if !widget_dir.exists() {
+        std::fs::create_dir_all(&widget_dir)
+            .expect("Failed to create widget storage directory");
```
Charlie-XIAO, line 16:
```suggestion
        std::fs::create_dir_all(&widget_dir)?;
```

Do you really mean to panic here?

src-tauri/src/widget_api/fs/apis.rs
```diff
@@ -0,0 +1,154 @@
...
+
+#[command]
+pub fn read_file<R: Runtime>(
+    app_handle: AppHandle<R>,
+    widget_id: String,
+    path: String,
+) -> Result<String, String> {
+    let file_path = get_file_path(&app_handle, &widget_id, &path);
+    std::fs::read_to_string(&file_path)
+        .map_err(|e| format!("Failed to read file '{}': {}", file_path.display(), e))
```
Charlie-XIAO, line 29: Since we are using `anyhow`, I think the better way is to do:

```suggestion
    std::fs::read_to_string(&file_path).context("xxx")?;
```

This wraps an additional context `"xxx"` outside the error of `read_to_string` and propagates it. The signature of this function would then need to change. Same comment for the other commands in this file.

src-tauri/src/widget_api/fs/apis.rs
```diff
@@ -0,0 +1,154 @@
...
+    let widget_dir = widget_base.join(widget_id).join("storage");
+    if !widget_dir.exists() {
+        std::fs::create_dir_all(&widget_dir)
+            .expect("Failed to create widget storage directory");
+    }
+    widget_dir.join(path)
+}
+
+#[command]
+pub fn read_file<R: Runtime>(
```
Charlie-XIAO, line 22: Would we want to stick to the protocol for our internal commands, i.e., the return type should be `CommandOut<T>`? The main reason I designed such a protocol is to have a more consistent behavior when the returned value is sent to the frontend.

Charlie-XIAO (commented):
Thanks for looking into the widget API framework! These look nice overall. I left a few comments, and please let me know if there is any confusion.

You may refer to `commands.rs` to see the "protocol" that I defined there. I would kind of prefer keeping widget API commands in consistent with our internal commands (as I also mentioned in a review comment). Please let me know if you have different opinions.

I haven't yet looked at the tests, because my proposed changes may affect the tests as well. I will review the tests when the main structure is finalized.

> Without tauri plugin, we must register all commands in a single `invoke_handler` when calling `tauri::Builder`, which is nearly impossible to maintain.

I don't really agree with this. Since they are under different modules, we will have

```rs
        .invoke_handler(generate_handler![
            commands::bundle_widget,
            commands::open_widget_base,
            commands::refresh_widget_collection,
            widget_api::fs::xxx,
            widget_api::fs::xxx,
            widget_api::some_other_module::xxx,
            widget_api::some_other_module::xxx,
        ])
```

which is still pretty maintainable. This being said, I'm okay with the current plugin approach, unless there are known degradations compared with direct invoking.

src-tauri/src/widget_api/fs/apis.rs
Xinyu-Li-123: Actually, I don't really understand why there should be a `CommandOut` to start with. Why can't we just let the tauri command returns a `Result<T, String>`? If we call the command in rust, we can unwrap it; if we invoke it in javascript, the return value is a `Promise<any>`, which can be handled with either then/catch or try/catch.

For example, we can define a tauri command `my_command` like this
```rust
#[tauri::command]
fn my_command() -> Result<String, String> {
    // Some logic that might succeed or fail
    if some_condition {
        Ok("Success message".to_string())
    } else {
        Err("Error message".to_string())
    }
}
```
and invoke it in javascript like this
```javascript
window.__TAURI__.invoke('my_command')
    .then((result) => {
        // This block will execute if the Rust side returns Ok(...)
        console.log('Success:', result);
    })
    .catch((error) => {
        // This block will execute if the Rust side returns Err(...)
        console.error('Error:', error);
    });
```
or using async / await syntax
```javascript
async function callMyCommand() {
    try {
        const result = await window.__TAURI__.invoke('my_command');
        console.log('Success:', result);
    } catch (error) {
        console.error('Error:', error);
    }
}

callMyCommand();
```

If the command returns a `CommandOut<T>`, the one who invoke it in javascript will need to explicitly check the status code, which will make the invocation more confusing, in my opinion
```
window.__TAURI__.invoke('my_command')
    .then((result) => {
        if (result.success) {
            console.log('Success:', result.success);
        } else if (result.failure) {
            console.error('Failure:', result.failure);
        }
    })
    .catch((error) => {
        // This block will execute if there's an unexpected error in the invocation itself
        // Not the same as a 'Failure' response from the command
        console.error('Error:', error);
    });
```

src-tauri/Cargo.toml
Xinyu-Li-123: Sorry, will fix this in the next commit

src-tauri/src/widget_api/dummy/apis.rs
Xinyu-Li-123: Will remove it in the next commit

src-tauri/src/widget_api/fs/apis.rs
Xinyu-Li-123: Will fix this in the next commit

src-tauri/src/widget_api/fs/apis.rs
Xinyu-Li-123: Got it. Will modify all error message to use `context`.

src-tauri/src/main.rs
Xinyu-Li-123: Got it. I will fix my comments. Also, I wonder if there is a general coding style standard for rust and javascript/typescript we should refer to. If so, could you provide the source to such standards?

src-tauri/src/main.rs
Xinyu-Li-123: Will remove it in the next commit.

src-tauri/src/widget_api/fs/apis.rs
Charlie-XIAO: Interesting, but I'm still -0.5 on this. I think this is unnecessary restriction, in that: (1) not forcing `storage` does not cause security issues, and (2) users can still follow such a structure to organize their widget directory, as long as they specify `storage/xxx` instead of `xxx` when invoking the API.

On the other hand, the API would be more flexible without this restriction. For instance, users can name the storage folder as they like (e.g., `assets/`). They can also have multiple storage directories (`public/`, `images/`, `posts/`) for different purposes without having to use a top-level folder. In rare cases users may want to read their own source code / README / etc. that are commonly in the root?

Xinyu-Li-123: Thak makes sense, and it will make my life easier as well. Let me remove the `storage` in later commits.

Xinyu-Li-123:
@Charlie-XIAO I am reopening this PR for review because I have implemented a working set file system widget api with access validation. Below is a list of changes
1. More FS Widget apis: 
    * read / write / append / delete file, 
    * is_file / is_dir / exists, 
    * create / remove dir
2. File system access with validation to ensure that the widget only has access to contents within widget directory
3. Use of `anyhow::Context` for better error message. To use this, we need non-tauri-command functions to return `Result<T, anyhow::Error>, and tauri commands to return `Result<T, tauri::InvokeError>`. The conversion from `anyhow::Error` to `tauri::InvokeError` is done using `tauri::InvokeError::from_anyhow`. An example is given below
   ```rust
    // non-tauri-command function. Used as util functions by tauri commands
    fn bar(flag: bool) -> AnyhowResult<String, AnyhowError> {
        if flag {
            Ok("OK".to_string())
        } else {
            Err(AnyhowError::msg("Not OK"))
        }
    }

    // tauri command
    #[command]
    pub fn foo<R: Runtime>(
        app_handle: AppHandle<R>,
        widget_id: String,
        flag: bool,
    ) -> Result<String, InvokeError> {
        bar(flag)
            .context("Failed to call bar in foo")
            .map_err(InvokeError::from_anyhow)
    }

   ```

src-tauri/src/widget_api/fs/utils.rs
```diff
@@ -0,0 +1,325 @@
...
+    // Setup a specific widget directory within the widget base environment
+    fn setup_widget_directory(widget_base: &PathBuf, widget_id: &str) -> PathBuf {
+        let widget_dir = widget_base.join(widget_id);
+        fs::create_dir_all(&widget_dir).expect("Failed to create widget directory");
+        widget_dir
+    }
+
+    #[test]
+    fn test_get_widget_base() {
+        let (app_handle, expected_base, _temp_dir) = setup_base_environment();
```
Charlie-XIAO, line 178: NIT I think `_` suffices if you do not use `_temp_dir`.

src-tauri/src/widget_api/fs/mod.rs
```diff
@@ -0,0 +1,24 @@
+mod apis;
+pub mod utils;
+
+use tauri::{
+    plugin::{Builder, TauriPlugin},
+    Runtime,
+};
+
+// The registered apis can be invoked as "plugin:widget_api.fs|<api_name>"
```
Charlie-XIAO, line 9:
```suggestion
/// The registered APIs can be invoked as `"plugin:widget_api.fs|<api_name>"`.
```

NITS, and since we do call this function I think we can use docstring style.

src-tauri/src/widget_api/fs/apis.rs
```diff
@@ -0,0 +1,486 @@
...
+    let file_path = utils::get_entry_path(&app_handle, &widget_id, &path);
+    if !file_path.is_file() {
+        return Err(InvokeError::from(format!(
+            "Path '{}' is not a file",
+            file_path.display()
+        )));
+    }
+
+    std::fs::read_to_string(&file_path)
+        .context(format!("Failed to read file '{}'", file_path.display()))
```
Charlie-XIAO, line 90: TIL, I'm using `path.to_string_lossy()` or `{:?}` elsewhere but `path.display()` looks much better. I think I would open a PR to switch when I have time.

src-tauri/src/widget_api/fs/apis.rs
```diff
@@ -0,0 +1,486 @@
...
+        .map_err(InvokeError::from_anyhow)?;
+    let folder_path = utils::get_entry_path(&app_handle, &widget_id, &path);
+    if folder_path.exists() {
+        return Err(InvokeError::from(format!(
+            "Directory '{}' already exists",
+            folder_path.display()
+        )));
+    }
+    std::fs::create_dir_all(&folder_path)
+    // .map_err(|e| {
```
Charlie-XIAO, line 168: Remove if not needed.

src-tauri/src/widget_api/fs/apis.rs
```diff
@@ -0,0 +1,486 @@
...
+    app_handle: AppHandle<R>,
+    widget_id: String,
+    path: String,
+) -> Result<(), InvokeError> {
+    utils::validate_entry_path(&app_handle, &widget_id, &path)
+        .context("Failed to validate entry path")
+        .map_err(InvokeError::from_anyhow)?;
+    let folder_path = utils::get_entry_path(&app_handle, &widget_id, &path);
+    std::fs::remove_dir_all(&folder_path)
+    // .map_err(|e| {
```
Charlie-XIAO, line 186: Remove if not needed.

Charlie-XIAO (commented):
Thanks for the impressive work. I left another round of suggestions, which mainly regards styles and code readability. Please understand that   if there are no fundamental errors, I need to first improve readability, consistency, and conciseness of codes, and then care about more detailed logics. Thanks for your patience.

And as you asked, here are some docstring standards that is widely adopted by the Rust community, though not within the capabilities of `rustfmt` (I will try to formulate when I have time):

#### Comments

These are the most comments that start with `//` (note we never use `/* */` in Rust though it is supported). No need to say, these are for readability purpose (and sometimes used to separate code into blocks, also for readability). The principles are:

- Exhaust each 88-character line: do not switch line early if the next word/component can still fit into the same line and is in the same paragraph.
- Do not exceed the 88-character limit. The only exceptions are URLs. Long codes may also exempt from this principle if it's ugly to write in multi-line format enclosed in triple backticks.
- Inline codes should always be enclosed in single backticks. There is argument around whether simple inline codes like `true` or `None` should be enclosed in backticks: I lean towards that they should.
- Prefer double quotes over single quotes whenever possible and reasonable.
- No need for any indentation within a paragraph (because I see you have this habit). This may cause some documentation renderers to render those texts inside `<quote></quote>`.

#### Docstrings

Docstrings that start with `///` are for functions/structs/traits/enums/etc. and `//!` are (top-level) for modules/files. In addition to the principles for normal comments, they further follow the following conventions:

- The first line should fit into one line, i.e. within 88 characters. This is called "short summary".
- If there are to be more details, they should be an empty line away from the summary.

src-tauri/src/widget_api/fs/utils.rs
```diff
@@ -0,0 +1,325 @@
...
+///
+/// Note that if the folder does not exist, the function will return an error, since this
+/// likely indicates a mistake in the widget ID.
+pub fn validate_widget_id<R: Runtime>(
+    app_handle: &AppHandle<R>,
+    widget_id: &str,
+) -> Result<(), AnyhowError> {
+    if cfg!(debug_assertions) {
+        println!("Validating widget ID: '{}'", widget_id);
+    }
```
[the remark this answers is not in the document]
Xinyu-Li-123: I'm not sure about that. These `println!` will only be printed in debugging mode. In our case, it's when we use `npm run tauri dev`. It does helps with my debugging, because I can inspect intermediate variable using these debugging prints. We may remove some of the debugging print if they seems unnecessary, but I think the general idea of "having debugging print in source code" should be adopted.

> Especially since we are releasing as an application instead of as a library.)

If I understand correctly, we will **release** the binary compiled for different platform instead of the source code, though people can clone from Github and build it themselves. If someone need to look into the code, they are likely experimenting with Deskulpt, in which case a more detailed debugging print will be helpful.

src-tauri/src/widget_api/fs/utils.rs
```diff
@@ -0,0 +1,325 @@
...
+    let widget_dir = get_widget_dir(app_handle, widget_id);
+    // if canonicalized() is used, and the app runs on Windows, the long path prefix "\\?\" will be added
+    //  to the path, which may cause issues with path comparisons.
+    let widget_dir_absolute = widget_dir.absolutize().map_err(|e| {
+        AnyhowError::msg(format!(
+            "Failed to get absolute path of widget directory: {}",
+            e
+        ))
+    })?;
+    if !widget_dir_absolute.exists() {
```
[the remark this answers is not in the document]
Xinyu-Li-123: Yes, but I think "ease of maintainence" matters as well, and supports for debugging will definitely make our (Deskulpt developers') life easier.

src-tauri/src/widget_api/fs/utils.rs
```diff
@@ -0,0 +1,325 @@
...
+                path
+            )));
+        }
+    }
+
+    Ok(())
+}
+
+/// Get the widget base directory from the app state, return the absolute path of the widget base directory.
+pub fn get_widget_base<R: Runtime>(app_handle: &AppHandle<R>) -> PathBuf {
```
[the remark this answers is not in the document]
Xinyu-Li-123: These simple functions are not meant to shorten the code. It serves to provide modularity and nice abstraction that hides implementation details. Using these one-liner utililty functions, I could reason about abstract things like "getting widget dir", or "getting file path", without worrying about its implementation details everytime.

Moreover, if we decide to determine widget dir differently later (e.g. reading from a Deskulpt app config file so that users can customize widget dir), I only need to change these utililty functions without modifying other functions that depends on them.

src-tauri/src/widget_api/fs/utils.rs
```diff
@@ -0,0 +1,325 @@
...
+/// Get the widget directory from the app state given the widget ID.
+pub fn get_widget_dir<R: Runtime>(
+    app_handle: &AppHandle<R>,
+    widget_id: &str,
+) -> PathBuf {
+    get_widget_base(app_handle).join(widget_id)
+}
+
+// Get the entry path in the widget directory given the widget ID and the entry path.
+pub fn get_entry_path<R: Runtime>(
```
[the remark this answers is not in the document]
Xinyu-Li-123: I was thinking about "file system entry" here, which includes both file and directory. The function is supposed to get the absolute path to a file or directory in the widget dir. But I see your point. It is confusing given we already have "entry point" in widget config. Let me change it to something else. 

> because I see you assigning let `file_path = get_entry_path(...)`

If you check the `create_dir` function, you will also see a `let folder_path = get_entry_path(...)`.

src-tauri/src/widget_api/fs/utils.rs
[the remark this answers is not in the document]
Charlie-XIAO: In short, this is the job of debuggers. In details, this is not maintainable code.

- Everyone has different variables they want to inspect, and when they are debugging different issues. And everyone can have valid reasons.
- What if I'm debugging some other issue that is simply unrelated to widget API? These debugging prints will mess up my console output.
- How do you determine to which extent we include debugging prints in the codebase? If "somewhere", why? If "everywhere", I cannot imagine the complexity of the codebase and the extra maintenance burden.

Finally, though I think this does not count as a valid reason, I don't see large Rust projects doing this way.

src-tauri/src/widget_api/fs/utils.rs
```diff
@@ -0,0 +1,325 @@
...
+    app_handle: &AppHandle<R>,
+    widget_id: &str,
+    path: &str,
+) -> PathBuf {
+    get_widget_dir(app_handle, widget_id).join(path)
+}
+
+#[cfg(test)]
```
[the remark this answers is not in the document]
Xinyu-Li-123: Yeah, I'm not gonna further work on them neither. I think I have covered common cases with these tests, and I'm good with that within this PR. Let's work on them in later PRs.

src-tauri/src/widget_api/fs/utils.rs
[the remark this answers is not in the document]
Xinyu-Li-123: Yeah, that makes sense. I should just use gdb (or other debugger) to do these jobs. And these are only messages I'm interested in for debugging purposes. Let me remove these lines in the next commit.

src-tauri/src/widget_api/fs/utils.rs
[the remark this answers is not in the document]
Charlie-XIAO: Well but you can stick to your point. I'm really just -0.5 and you can make the choice.

src-tauri/src/widget_api/fs/utils.rs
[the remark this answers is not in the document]
Charlie-XIAO: I think `resource` is nice :)

src-tauri/src/widget_api/fs/utils.rs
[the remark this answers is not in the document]
Charlie-XIAO: If that's the case, we can postpone tests to another MAINT PR as well.

We can explicitly link to that PR in the top comment so that when the TA looks at this PR he will find out where the tests are.

src-tauri/src/widget_api/fs/utils.rs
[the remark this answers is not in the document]
Xinyu-Li-123: I see your point, but I believe reusability is not the mere purpose of modularize code. Better readability is also an importatn motivation. So I will stick to using these two one-liner function. I believe the performance drop of calling an extra function is negligible for our application.

src-tauri/src/widget_api/fs/apis.rs
```diff
@@ -0,0 +1,486 @@
...
+//         .context("Failed to call bar in foo")
+//         .map_err(InvokeError::from_anyhow)
+// }
+
+#[command]
+pub fn exists<R: Runtime>(
+    app_handle: AppHandle<R>,
+    widget_id: String,
+    path: String,
+) -> Result<bool, InvokeError> {
```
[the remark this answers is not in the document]
Xinyu-Li-123: Actually, I'm not sure. I use InvokeError  because anyhow error can't be directly returned by a tauri command, and `from_anyhow` is the only method I can find to conver the anyhow error to a tauri-compatible one. But it does work in the frontend, at it display error message as expected in the developer console.

src-tauri/src/widget_api/fs/apis.rs
```diff
@@ -0,0 +1,486 @@
...
+use tauri::{command, AppHandle, InvokeError, Runtime};
+
+use crate::widget_api::fs::utils;
+
+use anyhow::Context;
+
+// TODO: (Future) Write formatted string to files (now there is no way to break new lines)
+// TODO: (Future) Write auto-generated unittests to cover more corner cases
+
+// fn bar(flag: bool) -> AnyhowResult<String, AnyhowError> {
```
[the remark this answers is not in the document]
Xinyu-Li-123: I will remove the foo/bar example. But I think todo items should be kept in the source code, even if we have a separate documents for them. What do you think?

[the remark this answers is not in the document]
Charlie-XIAO: Sure. (I actually think I didn't mean to include those two lines.)

Xinyu-Li-123:
@Charlie-XIAO Please check the updates.

Charlie-XIAO (commented):
Thanks for the update. I left some detailed minor comments in `apis.rs` and a general major comment in `utils.rs`.

Also regarding `apis.rs`, do you think we want a macro for generating those commands, given they all follow a similar pattern? For instance,

```rust
macro_rules! fs_command {
    ($fn_name:ident, $ok_type:ty, $expr:expr $(, $arg_name:ident : $arg_type:ty),* $(,)?) => {
        #[command]
        pub fn $fn_name<R: Runtime>(
            app_handle: AppHandle<R>,
            widget_id: String,
            path: String,
            $($arg_name: $arg_type),*
        ) -> Result<$ok_type, InvokeError> {
            utils::validate_resource_path(&app_handle, &widget_id, &path)
                .context("Failed to validate resource path")
                .map_err(InvokeError::from_anyhow)?;
            let file_path = utils::get_resource_path(&app_handle, &widget_id, &path);
            $expr(&file_path, $($arg_name),*)
        }
    };
}

// command name
// type of the returned ok variant
// a function that takes the validated file path and other arguments if any
// other arguments, optional

// Example 1, no other arguments
fs_command!(func1, bool, (|file_path: &PathBuf| Ok(file_path.exists())));

// Example 2, one other argument
fs_command!(
    func2,
    (),
    (|file_path: &PathBuf, content: String| {
        std::fs::write(file_path, content)
            .context("xxx")
            .map_err(InvokeError::from_anyhow)
    }),
    content: String,
);
```

But I'm not good at writing macros (at least not yet). The above does not seem to simplify code a lot. If we narrow down the cases to be covered by the macro maybe we can get something better. If we cannot think of better macros I think we may not need it for this one.

src-tauri/src/widget_api/fs/utils.rs
```diff
@@ -0,0 +1,275 @@
...
+/// Note that if the folder does not exist, the function will return an error, since this
+/// likely indicates a mistake in the widget ID.
+pub fn validate_widget_id<R: Runtime>(
+    app_handle: &AppHandle<R>,
+    widget_id: &str,
+) -> Result<(), Error> {
+    let widget_base = get_widget_base(app_handle);
+    let widget_dir = get_widget_dir(app_handle, widget_id);
+    // If canonicalized() is used, and the app runs on Windows, the long path prefix "\\?\" will be added
+    // to the path, which may cause issues with path comparisons.
```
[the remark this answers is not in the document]
Charlie-XIAO: Also, I think between the following two `if` clauses, the first one can be combined, and the second one can use the error message of the first one. Note that `is_dir` also checks that if the path exists IIRC. I think there is no need to make such a subtle distinguishment between the two cases.

src-tauri/src/widget_api/fs/utils.rs
```diff
@@ -0,0 +1,275 @@
+use crate::states::WidgetBaseDirectoryState;
```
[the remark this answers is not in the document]
Charlie-XIAO: Let's revisit in another PR.

Charlie-XIAO (approved):
I removed the PR and cleaned up some very minor issues. This is good to merge now. I will open another PR with your tests from this PR. Thanks for the impressive work, and looking forward to what you can build on top of this :)

Charlie-XIAO:
Merging @Xinyu-Li-123, congrats

PR #14, opened 2024-04-07
Widget api updated

rz2606:
Adapted from #12
In this PR, I add 'str' to the previous demo (two sets of widget apis: `dummy`, `fs`), but there's a minor problem to fix in the new plugin str, when the react component invokes a widget API, the return value is empty.

### 1. Access to Widget-specific Data
Widget apis are tauri commands exposed to widgets. Conceptually, to invoke a widget api, the api's implementation must have access to the widget id. For example, to read a file from the widget storage, the `get_text` api` need to know the widget's id. As a result, the function signature of `get_text` should be something like
```rust
#[command]
pub async fn get_text(widget_id: String) -> String {
    "Hello from Tauri!".to_string()
}
```
We would have a base `Widget` class like this
```typescript
class Widget: 
  // "#" means private variable
  #widget_id: string = "";
  
  constructor(widget_id: string) {
    this.#widget_id = widget_id;
  }

  // invocation of `get_text` api is wrapped in a method so that it has access to `widget_id`
  async get_text(path: string) {
    const result = await invoke("get_text", { widget_id: this.#widget_id});
    ...
  }

  // to be overridden by subclass widgets
  render() {
    throw new Error("Not implemented")
  }
```
A widget developer may extend this base widget class to define their own widget.

## Widget API

```
plugin:
   name: "widget_api.dummy"
   commands:
      get_text(widget_id: String) -> String

plugin:
    name: "widget_api.str"
    commands:
      // Get text and display it 
      get_text(app_handle: AppHandle, widget_id: String, path: String) -> Result<String, String>
```
As an example, to invoke `get_text` in Javascript, we can do this
```javascript
invoke("plugin:widdget_api.fs|get_text", {widget_id: "ok-poem-display"})
```

## How to Test this PR

Please first checkout the PR branch locally. For testing widgets, [this repo](https://github.com/CSCI-SHU-410-SE-Project/test-class-widgets) contains some simple widgets defined as React class components.


Charlie-XIAO:
Closing this one since it is no longer relevant.

Closing does not change the link so it will still be valid when the TA reviews our submission.

PR #17, opened 2024-04-08
FIX fix cargo lock, accidentally include version number of normpath when resolving merge conflict

Xinyu-Li-123:
As is described in the title. Only change one line of Cargo.lock, so I will force merge it.


Charlie-XIAO:
I think we shall investigate if there can be a workflow to auto-resolve lock file conflicts when merging into `main`. The thing is that `git` cannot resolve lock file conflicts well.

PR #23, opened 2024-04-12
ENH not removing deleted and ignored widgets on rescan

### LINKED ISSUE #6: BUG "Render All" button won't remove deleted / ignored widgets
From https://github.com/CSCI-SHU-410-SE-Project/Deskulpt/pull/3#issuecomment-2036444329 by @Xinyu-Li-123:

If a widget is rendered, and later deleted or ignored, it won't be cleared from the canvas. Below is what I have tested:
- Delete a rendered widget:
  - add widget to `$APPDATA/widgets`; 
  - click `rescan` and `render all`; 
  - remove the widget folder; 
  - click `rescan` and `render all`;
- Ignore a rendered widget: 
  - add widget to `$APPDATA/widgets`; 
  - click `rescan` and `render all`; 
  - change the `ignore` option in `deskulpt.conf.json` from `false` to `true`
  - click `rescan` and `render all`;

Charlie-XIAO:
Fixes #6.


ROMEEZHOU (commented):
This is not necessarily a bug but I noticed that when a new widget is added to the widget folder, user has to click "rescan" and then "render all" to show the widget on canvas. (however, when a widget is deleted, user has to only click rescan and the widget would be automatically removed from the canvas) Perhaps we can also make new-added widgets automatically appear on canvas.

Charlie-XIAO:
Good suggestion. The updated logic is:

- Scan through the widget base directory
- Remove deleted/ignored widgets from both the manager and the canvas
- Add created/unignored widgets to the manager and auto-render them on the canvas; note that widgets that are already on the screen will not be automatically re-rendered (i.e., not equivalent to "Render All")

ROMEEZHOU (approved):
LGTM

PR #25, opened 2024-04-12
TST refine widget `fs` API unit tests

Charlie-XIAO:
Continuation of #13.


Xinyu-Li-123:
@Charlie-XIAO I have refactored the unit test for fs api. Could you write a review for it?

Charlie-XIAO:
LGTM, here is the coverage report:

image

I have just noticed that it was me who opened the PR, so I cannot approve it myself. However, the main works are done by @Xinyu-Li-123 and I just modified on top of that, so I will override the requirements and merge (especially due to the deadline).

PR #15, opened 2024-04-08
TST unit tests for `config.rs` and `testing` module for convenience

Charlie-XIAO:
This PR creates tests for `config.rs`, as described in the title.

It also creates the `testing.rs` module which currently contains the `assert_err_eq` utility to compare against an error against a `anyhow` error chain, which can be used by future unit tests as well.

PR #26, opened 2024-04-13
TST/FIX fix bundler tests failures on Windows

Charlie-XIAO:
The bundler tests introduced in #10 failed on Windows, blocking the cross-platform testing CI to be set up.

This PR does the following:

- Updates the logic of rejecting relative imports beyond the root. The previous version is incorrect on non-Windows platforms because we made a simple `join` operation that does not clean up `..` components.
- Improves error message display (prefer `.display()` over `{:?}`, put single quotes around paths).
- Replace `parameterize` with `rstest`, as introduced in #15.
- Replace `assert_err_eq` by the one from the `testing` module, as introduced in #15.
- Separated the absolute import test case, because the path of the absolute import need to be determined at runtime and is different across platforms.
- Minor cleanup of the fixtures (including renaming).

I have tested on Windows and Linux (WSL). @ROMEEZHOU would be nice if you can give a review and test on macOS since the test workflow is not ready yet.


rz2606:
How does Tauri widget support third-party component libraries? Such as bootstrap-react.

Charlie-XIAO:
Do you mean "external dependencies"? How is the question related to this PR? @rz2606

ROMEEZHOU (approved):
LGTM :)

PR #28, opened 2024-04-13
MAINT add pull request labeler

Charlie-XIAO:


Charlie-XIAO:
It seems that `pull_request_target` workflows are simply not triggered when it is first created in a PR (i.e., without existing on `main`). We would have to merge and see. This is not big deal so I will force the merge.

PR #29, opened 2024-04-13
ENH support TypeScript (including `tsx`) in bundler

Charlie-XIAO:
This PR implements TypeScript support, as described in the title. It is overall simple: the only point to notice is `verbatim_module_syntax: true` in the configuration which avoid removing unused imports from the module (we need to keep `React` in scope). See [TypeScript documentation](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax) for more information about this configuration. Tree-shaking should be able to handle other unused imports afterwards.

Note that there are some issues with reading `package.json` in `config.rs` at this moment. In particular if one has a `package.json` without the `dependencies` field the manager will be errored out. It causes some difficulties in testing because somehow we may want to install `@deskulpt-test/react` and `typescript` as `devDependencies` and leave `dependencies` empty. I will open another PR to fix the problem since it will involve updating tests and commands.


Charlie-XIAO:
This PR is mostly good: the thing we are missing here is tests. Just posting this comment to avoid myself forgetting this.

Charlie-XIAO:
Since we are hurrying for the deadline I'm going to self-merge this one as well. This provides only very basic TypeScript support, and we may have more to concern regarding TypeScript and the bundler in the future:

- Unit tests, of course.
- Is the currently solution able to strip away type imports, or we have to write `import type xxx` to let the bundler know?
- How are we possibly going to deal with `@emotion/react`? This is hard regarding both React and TypeScript.
- Can we automatically add `import React from xxx` for `jsx` and `tsx` without requiring users to specify?
- Are we going to support named imports from `@deskulpt-test/react`, perhaps by exporting a subset?
- Is it possible to allow named imports from `@deskulpt-test/apis`? This may be harder than the one above.
- etc.

PR #27, opened 2024-04-13
FEAT Make each widget div draggable

rz2606:
Successfully integrate Xinyu's Deskulpt-demo into our Deskulpt project to adjust canvas layout.
**Reference:**
https://jsfiddle.net/wfbY8/4/

**Dependencies:** 
Add window dependency WIN32
Tauri's dependency set to 1.6.1

**Key improvements:**
Make the window fullscreen, transparent
Make each widget div draggable

**How to use:**
1. Use keyboard shortcut ESC to SINK CANVAS. Widgets can not be manipulated in this situation.
image

2. Press FLOAT CANVAS on Deskulpt manager panel to have the canvas displayed above any windows.
image

**Further work:** 
1. Keep canvas below desktop icons.
2. Experiment cross-platform

PR #31, opened 2024-04-14
ENH make widget `<div>` draggable

rz2606:
An updated version of #27 to make div draggable.
(Import Draggable component from the "react-draggable" package)


src/components/WidgetContainer/index.tsx
```diff
@@ -1,7 +1,7 @@
 import { Box, Tooltip } from "@mui/material";
 import InfoIcon from "@mui/icons-material/Info";
 import React from "react";
-
+import Draggable from "react-draggable";
```
Charlie-XIAO, line 4:
```suggestion
import Draggable from "react-draggable";

```

Please leave an empty line after import statements.

src/components/WidgetContainer/index.tsx
```diff
@@ -19,29 +19,30 @@ export default function WidgetContainer(props: {
...
-        {inner}
-      </Box>
+      <Draggable>
+        <Box
+          sx={{
+            px: 2,
+            py: 1,
+            m: 1,
+            borderRadius: 1,
+            // bgcolor: "lightblue",
```
Charlie-XIAO, line 29: Is there a reason for removing this? If you think the lightblue background color is too distracting you may also try the following:

```suggestion
            border: "2px solid black",
            bgcolor: "rgba(0, 0, 0, 0.2)",
```

src/components/WidgetContainer/index.tsx
```diff
@@ -19,29 +19,30 @@ export default function WidgetContainer(props: {
...
+            position: "relative",
```
Charlie-XIAO, line 30:
```suggestion
```

I think with the draggable wrapper, we can remove this one.

Charlie-XIAO (commented):
Thanks @rz2606, this one looks nicer! I've let some minor comments. LGTM once those are resolved :)

Charlie-XIAO (approved):
I've pushed the suggestions. Merging on CI green.

PR #35, opened 2024-04-16
FEAT Frontend wrapper for `@deskulpt/apis`

Xinyu-Li-123:
In this PR, I have implemented the frontend wrapper for default dependencies. Default dependencies are dependencies provided to widgets by Deskulpt. The wrapper covers two types of default dependencies
- those that are shared by all widgets, such as `React.useState()`
- those that are widget-specific, such as `readFile()`
 
Below is a brief example of widget source code that uses these two types of dependencies. For a full example, see [this writer widget](https://github.com/CSCI-SHU-410-SE-Project/test-widgets/tree/main/ok-writer)
```typescript
import React from "@deskulpt/react"
import apis from "@deskulpt/apis"

console.log(React.useState)
console.log(apis.fs.readFile)
```

## Default depedencies shared by all widgets

These are dependencies shared by all widgets (and by Deskulpt manager). For now, I only wrap the default export `React` of react. A widget may use the react from default dependencies like this
```typescript
import React from "@deskulpt/react"

console.log(React.useState)
```
This is achieved by several level of indirections.

First, we have a wrapper of react in Deskulpt source code
```typescript
// src/@deskulpt/react.index.ts
import React from "react";
export default React;
```

To make this wrapper module accessible to the webpage in canvas window (and thus to all widgets in that window) when Deskulpt is running, we do two things

1. When building Deskulpt, we let rollup preserve module structure by setting `preserveModule` to true in vite config's rollup options. This way, rollup won't bundle the source code. Instead, it will only convert node module import to relative path import, and file structure won't be changed after building [^1]. The problem with bundling is that variables and functions will be renamed and module structure will change. For example, rollup may convert `export default React` in our source code to `export { React as R }` in the bundle.

	When Deskulpt is running, the source code for our wrapper of react is configured to be stored at `/assets/src/@deskulpt/react/index.js`, and js code running in the browser can import from it via `import React from "/assets/src/@deskulpt/react/index.js"`

2. To redirect import from `@deskulpt/react` to `/assets/src/@deskulpt/react/index.js`, we create an importmap in the canvas webpage [^2]
   ```json
	{
	"imports": {
		"@deskulpt/react": "/assets/src/@deskulpt/react/index.js",
		}
	}
   ```

## Default dependencies that are widget-specific

In Deskulpt, these types of deps are mainly widget apis. For example, consider the widget api `async function readFile(path: string): string` that reads a file from the widget's folder and returns the content as string. A call of `readFile("output.txt")` by different widgets should read from `output.txt` in different widget folders. 

Roughly speaking, this is implemented by letting different widgets import different `readFile` that have different `widgetId` "associated" to it. 

### First Attempt

Consider two widgets with ids `todolist` and `calendar`. A first attempt would be to let widget developer explicitly import from different default dependencies.

For `todolist`, we have
```typescript
// $APPDATA/widgets/todolist/src/App.tsx

import React from "@deskulpt/react"
import apis from "todolist/@deskulpt/apis"

const content = await apis.fs.readFile("output.txt")
// Equivalent to 
// window.__TAURI_INVOKE__({
//	cmd: "plugin:widget_api.fs|read_file", 
//  widgetId: "todolist", 
//  path: "output.txt"
// })
```

For `calendar`, we have
```typescript
// $APPDATA/widgets/calendar/src/App.tsx

import React from "@deskulpt/react"
import apis from "calendar/@deskulpt/apis"

const content = await apis.fs.readFile("output.txt")
// Equivalent to 
// window.__TAURI_INVOKE__({
//	cmd: "plugin:widget_api.fs|read_file", 
//  widgetId: "calendar", 
//  path: "output.txt"
// })
// Note that the widgetId in the argument is different
```

The `todolist/@deskulpt/apis` should points to a javascript file like this (similar for `calendar`)
```javascript
// Source code of `todolist/@deskulpt/apis` module
// This is file is created at the runtime of Deskulpt so that it can 
// dynamically adapts to different widget ids

// Import from the raw apis that require explicitly passing in widgetId
import * as rawApis from "@deskulpt/apis"

let apis = {}
// Create a readFile function that calls the raw readFile api and pass in the widgetId
apis["fs"]["readFile"] = (...args) => rawApis.fs.readFile("todolist", ...args)
// ... wrapper for other raw widget apis

exports default apis;
```

This will work, but it has a catch: the widget developer need to explicitly write out the widget id in their source code. This is undesirable because it exposes too much details to widget developers, making the widget development error-prone (e.g. a widget dev may misspell the widget id). The widget id is the name of the widget folder in `$APPDATA/widgets`, and it should be managed by the Deskulpt manager.

### Solution: Redirection of import during widget bundling

What we want is that different widgets could import different apis by the same line of code below

```typescript
import apis from "@deskulpt/apis"
```

This is possible because we are in full control of the widget compilation. We can create the widget-specific import module and swap it in during widget compilation. This is achieved in two steps

1. Right before Deskulpt manager (javascript) invoke `bundle_widget`, we create a blob object of javascript code of the `todolist/@deskulpt/apis` module, and pass the url of the blob as an argument to `bundle_widget`.
2. In Rust backend, the bundler locates import statements whose source is `"@deskulpt/apis"`, and swaps it with the blob url.

This way, the widget developer doesn't need to explicitly specify the widget id in their source code. Instead, they can import from the same `@deskulpt/apis` module for different widgets, and Deskulpt will convert it to imports of widget-specific apis.

## How to test this PR

Clone the PR, install npm package, and load the [test widget](https://github.com/CSCI-SHU-410-SE-Project/test-widgets/tree/main/ok-writer) to see if the wrapped react and widget apis work.

[^1]: Note that minification, tree-shaking and other steps won't be affected. The only problem with this approach is that there won't be a single js bundle but multiple small js files. This could slow down the initial loading time of the webpage in the canvas window because of the overhead of requesting multiple files, but the slowdown is near negligible since all files are local and requests are based on IPC.

[^2]: When debugging, the files won't be bundled, and the wrapper is located at `/src/@deskulpt/react/index.ts` instead of `/assets/src/@deskulpt/react/index.js`. Since importmap can't be dynamically updated when the webpage is loaded, we create two canvas webpage html `canvs-build.html` and `canvas-dev.html`, and let npm choose which one should be the `cavnas.html` (see the `predev` and `prebuild` npm commands)


Charlie-XIAO:
Works fine on my Windows 11 machine, but got the following error on Ubuntu (WSL):

image

```jsx
import React from "@deskulpt/react";

function Counter() {
  const [count, setCount] = React.useState(0);

  return (
    <div>
      <div>Count: {count}</div>
      <button onClick={() => setCount(count + 1)}>Increment</button>
      <button onClick={() => setCount(0)}>Reset</button>
    </div>
  )
}

export default {
  render: () => <Counter />,
};
```

Xinyu-Li-123:
> Works fine on my Windows 11 machine, but got the following error on Ubuntu (WSL):

Can the [writer widget](https://github.com/CSCI-SHU-410-SE-Project/test-widgets/tree/main/ok-writer) import react normally?

Charlie-XIAO:
> Can the [writer widget](https://github.com/CSCI-SHU-410-SE-Project/test-widgets/tree/main/ok-writer) import react normally?

No luck.

> Can you also use devools to check if there is an importmap at the head of the canvas webpage? The import from `@deskulpt/react` should be redirected to an import from an absolute path `assets/src/@deskulpt/react/index.js`.

The importmap is there.

image

Xinyu-Li-123:
Sorry, I accidentally edit your comment instead of posting a new one. I have revert it back, mostly.

Could you try building the app instead? I think this may be due to my misunderstanding of how vite work. The build should work just fiine.

> The importmap is there, but it's pointing to `.ts` which is strange.

That is actually intentional. At least on Windows, when I use `npm run tauri dev`, the source code of Deskulpt won't be compiled because we are using a vite dev server. I find that I need to import from a different path and a different file, namely `/src/@deskulpt/apis/index.ts`. If you check out `package.json`, you will find that there is a predev and prebuild command that modify the `view/canvas.html` file according to whether we are developing or building Deskulpt, which essentially chooses different importmaps in the `view/canvas.html` for development and build.

Charlie-XIAO:
Same error in debug build. Here is the screenshot:

image

ROMEEZHOU (commented):
Works well on my machine

rz2606:
Also works well on my machine(Windows 11)
image

Charlie-XIAO:
@Xinyu-Li-123 Me and @ROMEEZHOU have been working on publishing `@deskulpt/...` packages and we have a somehow different architecture around this. This PR relied on `vite` for bundling, which (1) makes package publishing hard since files are messed together, and (2) has to distinguish between production and development builds. We alternatively plan to make the following:

```
root/
|-- src/
    |-- .scripts/   <-- Stores generated files for internal use
    |-- ..
|-- tooling/
    |-- react/
        |-- dist/   <-- Stores generated files for publishing
        |-- src/    <-- Source files, in this case `index.ts` which is a simple re-export of `React`
        |-- rollup.config.ts, package.json, tsconfig.json, etc.
    |-- ...
```

Using `rollup` for bundling, we would (1) generate ESM scripts and place under `/src/.scripts/` for internal use, and (2) generate CJS (or ESM) scripts and `.d.ts` type declaration files and place under `/tooling/react/dist/` for publishing. The import map in `views/canvas.html` would then be something like:

```
    <script type="importmap">
      {
        "imports": {
          "@deskulpt/react": "../src/.scripts/react.js"
        }
      }
    </script>
```

@ROMEEZHOU will likely make a PR for this tomorrow (only for `@deskulpt/react`). We can then quickly review that PR which should determine if the structure proposed above is desired. Then we can update the widget API part in this PR accordingly (and there will be fewer changes involved with that PR merged). WDYT @Xinyu-Li-123?

Xinyu-Li-123:
Maybe I can postpone review on my PR because I am going to need to integrate with the npm package you mentioned.

Xinyu-Li-123:
Also, I don't understand how should `react` or other default dependencies be available to the widget in the final  build. I notice you put an importmap at the head of canvas webpage when running `tauri dev`, but there is no such importmap when running with `tauri build`.

Xinyu-Li-123:
I have finally finished formating widget apis into npm package in the `/tooling` folder. The update contains two new npm packages
- `@deskulpt-test/raw-apis`
    This is the wrapper for raw apis. Widget id needs to be explicitly passed in to api calls. For example, `exists("ok-clock", "output.txt")`. This npm package is only used internally by `@deskulpt-test/apis`. It's not supposed to be published.
- `@deskulpt-test/apis`
    This is the wrapper for template fo widget-specific apis. It imports from `@deskulpt-test/apis` and apply widget id to the first argument of each api function. The output is stored in `/public/apis.txt`. At Deskulpt runtime, `/apis.txt` is fetched and widget id placeholder `${widgetId}` is replaced with the actual widget id. This allows us to generate widget-specific api module at runtime while providing valid typing support in a distributable npm package.

This version works, but it's quite hacky. So there are some catches (which I think could be fixed in later PR)

1. For the `@deskulpt-test/apis`, I can't find a better way to do the job: if compiled into js file
    - I can't fetch it correctly due to how Vite serve js files when running `tauri dev`; 
    - I can't fetch it correctly either when running `tauri build` since the txt file won't be included in the `/dist/assets/src/.script` folder. 

2. I have disabled the linting for `@deskulpt-test/apis` since this npm package because it requires some variables to be of `any` type.

3. Also, I didn't configure the publishing of these two packages. For now, you can install it locally by `npm install <local-path-to-npm-packages>`. I have tested it on my machine, and the tyepscript hint works if I manually install `@deskulpt-test/apis` in the widget dir this way.

## To Test this PR

Clone the PR, build in the following order

```
cd tooling/raw-apis
npm install
npm build

cd ../apis
npm install
npm build
```

To test if the linting works, install the `@deskulpt-test/apis` locally by

```
npm install <local path to the @deskulpt-test/apis package>

Xinyu-Li-123:
I have published `@deskulpt-test/raw-apis` as well. I find that I can't bundle the type declaration of `@deskulpt-test/apis` without installing `@deskulpt-test/raw-apis` because the bundled declaration involves a dynamic import of `@deskulpt-test/raw-apis`. I will have to install `@deskulpt-test/raw-apis` when installing `@deskulpt-test/apis` for correct typing support. Also, my previous published `@deskulpt-test/apis` has a relative path to `@deskulpt-test/raw-apis` in the `package.json`, which won't work when published. 

A direct consequence is that the node resolving of `@deskulpt-test/raw-apis` in rollup config of `@deskulpt-test/apis` is unnecessary, so I remove it as well.

Charlie-XIAO:
Thanks for doing all these heavy lifting. I will take a look later today.

Charlie-XIAO:
Pushed some stuff related to NPM packages. I still need to reorganize some source code to make things work again.

Charlie-XIAO (approved):
Major changes I pushed:

- Refactored the way we pass widget APIs blob URL, mainly in [`b3c640b`](https://github.com/CSCI-SHU-410-SE-Project/Deskulpt/pull/35/commits/b3c640b6365520a3d46ed75073e86c685663fd5c). This is due to semantic concerns.
- Refactored the way we manage states. I made it clearer what states `App` keeps, i.e., a shared part that will be synced with backend at times, and a frontend part that is purely managed by the frontend and does not need to bother the backend.
- Added blob URL revoke logic to avoid memory leak.

Thanks again for working out the way of exposing widget-specific APIs to users @Xinyu-Li-123. Approving and it's up to you whether to merge it as is :)

PR #40, opened 2024-04-21
TST update bundler test to use `@deskulpt-test/react`

Charlie-XIAO:
Follow-up of #37, in which we replaced `const React = window.__DESKULPT__.defaultDeps.React` with `import React from "@deskulpt-test/react"`. This PR updates bundler tests by using the new syntax, and also removes the global `__DESKULPT__` object from the canvas window.


ROMEEZHOU (approved):
LGTM :)

PR #41, opened 2024-04-22
FIX configuration error when `package.json` does not have `dependencies`

Charlie-XIAO:
If one has a `package.json` without the `dependencies` field the manager will be errored out. This PR fixes the issue and updates the test.


Charlie-XIAO:
I'm going to treat this as an emergency bugfix and self-merge on CI green. According to our recommended workflow for widget developers, they will install `@deskulpt/xxx` packages as dev dependencies but they may not install any dependencies (that will be treated as external dependencies). Failure to deal with a `package.json` file without `dependencies` is fatal.

I also did a bit of refactoring regarding `serde` renaming and top-level `use` statements.

Charlie-XIAO:
Shouldn't self-merge in principle, but we are lacking in time.

PR #42, opened 2024-04-23
ENH make config error fail only specific widgets

Charlie-XIAO:
Previously failure to load a widget configuration will fail the whole manager. For instance, if a `deskulpt.conf.json` file is missing the `name` field, the manager will not be able to scan any widget. The `refresh_widget_collection` command (which we use to rescan the widget base directory) now only fails if:

- There is an error reading the widget base directory.
- There is an error getting some entry in the widget base directory.
- There is an error inferring the widget ID from the path of the entry.

All of the three cases above are such that we cannot get the widget ID, thus we have no proper place to report for now (maybe in the future we will write logs). Failure to load a widget configuration will not lead to an overall failure of the command. Instead, the frontend will get the error message, though it will not be displayed for now (in the future we will have the information panel, which in normal case will show some configuration info and in error case will show the error information).


rz2606:
It works fine. When I delete the name field in the deskulpt.conf.json file of a widget(eg. counter), the test output is "???" for the counter entry. The changes in this PR improve the error-handling logic so that it will not lead to an overall failure of the refresh_widget_collection command.
 
In this branch, after changes, the frontend can receive the error message and have "???" displayed for the state widget, as shown below.

image

In the main branch, failure to load a widget configuration will lead to an overall failure of the refresh_widget_collection command. 

image
image

rz2606 (approved):

PR #45, opened 2024-04-24
RFC frontend widget state management

Charlie-XIAO:
Extracted from #35 to make it easier to review. This is also related to #23 regarding the cleanup logic part.


Charlie-XIAO:
Self-merging on CI green because this trivial and intermediate.

PR #46, opened 2024-04-24
ENH better message when bundling entry does not exist

Charlie-XIAO:
Try specifying an incorrect `entry` in `deskulpt.conf.json`, in `main` branch you will get something like "failed to canonicalize entry" which is not helpful. This PR provides a clearer error message.


ROMEEZHOU (approved):
LGTM :)

PR #47, opened 2024-04-26
FEAT/RFC/FIX major refactoring for final presentation

Charlie-XIAO:
There is no sufficient time to write a comprehensive description for this PR. Related will be presented in the final presentation and final deliverable. The size of this PR is also out of urgency for the deadline, otherwise it should have been split into ~10 smaller PRs.


Charlie-XIAO:
Self-merging as we need this for the final presentation demo.

PR #54, opened 2024-05-12
DOC generate frontend documentation

Xinyu-Li-123:
Add TypeScript doc generation via TypeDoc and integrate it into Docusaurus build workflow


Xinyu-Li-123:
@Charlie-XIAO Could you try build the website locally (the task with the name `Build TypeScript docs` in the github workflow), and see if the outcome is as expected?

Xinyu-Li-123:
@Charlie-XIAO I have migrated to the docusaurus plugin [typedoc-plugin-markdown](https://typedoc-plugin-markdown.org/). The typescript doc should work whtin the SPA generated by docusaurus.

Charlie-XIAO (approved):
Thanks for figuring out @Xinyu-Li-123, indeed this looks like a much nicer integration. I made some modifications on top of your work.

- You can just ignore changes in `src/`; they are caused by removing the `src/types/index.ts` files which is essentially re-exports and does not look well with the current `typedoc` configuration.
- I moved frontend and backend references to under a `Development` tab so that the top navbar is cleaner.
- The backend docs is referenced by absolute URL. Not sure if you agree, but my reasons are: (1) it is completely independent from docusaurus, and I have provided the `npm run rustdoc` command under `website/` so that it can be developed and built independently; (2) I do not want to bother with the routing issues; they seem to happen only to some machines, but they do happen, and we don't have a solution; (3) this is consistent with what we're doing in the workflow, i.e., we copy to the build directory directly instead of copying to the static directory and letting Docusaurus deal with them afterwards. Please let me know if you have different opinions.

And now that the references have been set up, there do seem to be modules and files that lack detailed docstrings, but I believe these should be taken care of in further PRs.

PR #61, opened 2024-06-05
RFC Refactoring of Deskulpt frontend UI

Charlie-XIAO:
This main purpose of this PR is to redesign the UI of Deskulpt.

### Choice of UI library

- Material UI: It is primarily designed for mobile and its components just looks like an Android app. Customizing it to achieve a modern desktop app look seems to be hard.
- Ant Design: Nice-looking components, powerful functionality. Yet it is too opinionated and hard to customize, and some components are too bloated with not-so-friendly API.
- Radix UI Primitives: Low-level primitives that are style-free. However, we are not a component library and it is unnecessary to do everything from the start.
- Shadcn UI: Built on top of Radix UI primitives. Looks very nice and is very popular nowadays, fully customizable because it simply let's you copy-paste its code. But again, we are not a component library and may not want to keep so many copy-pasted code in our code base. Moreover, see the next option.
- Radix UI Themes: **This is my choice for this PR.** Again built on top of Radix UI primitives, it looks very nice and modern especially in dark theme, as Shadcn UI is. Its default styling is based on pure CSS and CSS variables. Its design system is very nice is a sense that, if you look into this PR, there are not many places that I'm overriding styles with CSS-in-JS. Another benefit is that, when widget developers write their widgets they can use those CSS variables to get a consistent look with our app.

### Choice of styling

- CSS module: It's just... I don't like having an additional file for each component.
- Native CSS-in-JS: This is the `styles` prop. It's main limitation is that one cannot write nested styles such as `"&:hover"`.
- Emotion: **This is my choice for this PR.** It provides the `css` prop that solves the problem above.
- Styled components: Similar to CSS module, CSS and JS is not in one place which does not look intuitive to me.

### Provisioning widget UI

We can partially export `@radix-ui/themes` for widget developers, maybe designing some frequently used components ourselves as well. This would be very easy to learn, because most case people just wants to customize their margins, paddings, maybe flex gap, align, justify, etc. These are just extremely easy to do with the `Box` and `Flex` components without really writing CSS in JS. As mentioned before, one will also be able to use the CSS variables from Radix UI Themes to get the similar look to our app, for instance, `var(--space-2)`, `var(--red-1)`, `var(--radius-full)`. Note that these are also reactive to the theme so if users are using CSS variables for colors their components will a automatically support dark/light theme. These are also with almost no learning curve. For CSS-in-JS, we can export partially export `@emotion/react`. Then we need to find a way to let our bundler be able to compile things correctly (since `@emotion/react` is opinionated on how to interpret JSX syntax), but this should be doable. There is an existing emotion plugin for SWC; though we cannot use it directly we should be able to look into its source code and figure things out.

### Other comments

The padding of each widget container is removed. Hence one may find the previously written widgets too close to the border.


Charlie-XIAO:
I can now confirm that it is possible for our bundler/compiler to support the `@emotion/react` JSX runtime, so that we don't even need to import the sometimes unused `React` variable to make things work. Also, we can natively get the `css` prop as I have previously described which is similar to `style` but supports nested selectors so essentially most of the things (if not everything) one can do with CSS can be done with that prop. It mitigates the need to write a separate CSS file for widget developers, and I believe it is also the simpler way we support nested styling (just tweak the JSX transform and leave everything else to the JSX runtime). Perhaps it also leads to smaller app size as we do not need the CSS transforms part of the SWC crate (hopefully). The emotion babel plugin as I mentioned before, is not mandatory for later versions of `@emotion/react` and it only performs some optimizations rather than anything fundamental/functional, so that we simply don't need to bother ourselves especially given that we expect each widget to be "small".

Charlie-XIAO:
Everything's passing, so I'm merging anyways :) Feel free to leave comments here later or open new issues.

PR #62, opened 2024-06-07
ENH/FEAT Introduce `emotion`, JSX runtime, and UI library

Charlie-XIAO:
### `@deskulpt-test/ui`

This is mainly a partial re-export of `@radix-ui/themes` (we may include something self-defined components in the future if necessary). The reasons for not including some of theme is listed in the source file. Since `@radix-ui/themes` components are just adding a styling layer on top of `@radix-ui/primitives` (which are small), and we are using it for the UI of our app, and after all the style sheet is there, I believe this is the best choice (for now) for the UI of widgets.

### `@emotion/react`

`@emotion/react` is going to be the styling library we support for widget developers. The reasons, as I mentioned in #61, are:

- Compared with CSS modules: easier to support (see `bundler.rs`), CSS-in-JS is more intuitive
- Compared with `style` prop: supports nested selectors e.g. `&:hover`
- Compared with styled components libraries: more intuitive; styled components is overkill and has learning curve

Beside the partial re-export of `@radix-ui/themes`, the `@deskulpt-test/ui` package also exports `css` and `keyframes` from `@emotion/react`. These IMO are the only useful things for writing widgets. The following are the other things available from `@emotion/react` but I decide not to re-include:

- `jsx` and `createElement`: These IIRC are used for JSX pragma that configures the babel plugin and has nothing to do with our scenario
- `Globals`: This can add global styles. I don't think this is the recommended way to do CSS-in-JS. Also this is most commonly used for global styles like font faces, which is not really useful in our scenario since we should recommend users to use the CSS variables defines as in `@radix-ui/themes` available across our app.
- `ClassNames`: This supports passing CSS to some child element (in a strange way). I believe if someone *has* to do this, he should rethink whether he is using the correct component. This also IMO is something hard to understand.
- `ThemeContext`, `useTheme`, `ThemeProvider`, and `withTheme`: These are all used for theming. However widgets should rely on our native support for canvas theming e.g. dark/light mode by using the CSS variables.

### JSX runtime

Previously we have been using the classic JSX transform that requires to bring `React` into scope, yet the decision of supporting `@emotion/react` made me rethink the possibility of using the automatic transform. The [automatic JSX transform](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) imports `jsx`, `jsxs`, and `Fragment` from `react/jsx-runtime` which is automatically done without requiring to bring `React` into scope. Note that this new transform is [forced in React 19](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#new-jsx-transform-is-now-required) so sooner or later we need to adapt it.

On top of that, we need to support `@emotion/react/jsx-runtime` rather than just `react/jsx-runtime`. This is because we need the `css` prop on the components, so it matters how JSX syntax gets interpreted. In particular, `jsx` from `@emotion/react/jsx-runtime` will know how to deal with the `css` prop but `jsx` from `react/jsx-runtime` will not.

- `tooling/emotion/src/jsx-runtime.js`: This re-exports `@emotion/react/jsx-runtime` and is built under `src/.scripts/jsx-runtime.js` for internal use.
- In the bundler, specify to use the automatic runtime with import source `@deskulpt-test/emotion`. This will make the automatic JSX transform import `jsx`, `jsxs`, and `Fragment` from `@deskulpt-test/emotion/jsx-runtime`. Specified in `vite.config.ts`, this will then be mapped to `src/.scripts/jsx-runtime.js`.

### Other changes

- Update the test suite; reviewer may ignore `src-tauri/tests/fixtures/` in the first pass of review because there are plenty of changes due to the new JSX runtime and refactoring
- Add `globals` to `devDependencies` of each tooling package (it is used in rollup configuration)


Charlie-XIAO:
I've iterated and reviewed my code several times and I think it is good to go.

- The adoption of JSX runtime is pure improvement without doubt.
- The new tooling package does not add much to size of the app since most components we are already using in the main app and they are indeed pretty small.
- Modifications to the tests are not *removing* and previously supported stuff but only addition and simplification, and the tests are passing.

I'm just going to ping @ROMEEZHOU for a review to confirm that this works on Mac.

ROMEEZHOU (commented):
The core logic works well! The only thing is, on Mac, the "open base folder" button on the manager window gets outside the window.

Charlie-XIAO:
I believe the problem is not caused by this PR but by #61. I will open a follow-up PR to fix.

ROMEEZHOU (approved):
Then I don't see any problem specifically related to this PR : )

Charlie-XIAO:
See #65 ;)

PR #63, opened 2024-06-08
ENH named exports from `@deskulpt-test/react`

Charlie-XIAO:
This PR aims to best simulate the current behavior of react. This is both for better user experience, but also for our future work (e.g. external dependencies).

One more thing: This PR specifies `react` and `@types/react` as `peerDependencies`. With `npm` later than version 7 `peerDependencies` will be automatically installed as `dependencies` are, and on top of that `npm install` will complain if versions mismatch (which is desired), if I understand correctly.


Charlie-XIAO:
Merging because this is not harming anything.

PR #64, opened 2024-06-08
FIX error in the exported render function breaks react render loop

Charlie-XIAO:
Try doing some nonsense in the exported `render` function, e.g.

```js
export default {
  render: () => nonExistentObject,
  width: "300px",
  height: "100px",
};
```

This will in fact fail the render of all widgets and show no error on the canvas. This is because in the current implementation the `render` function is called within the state setter using a updater function, which will be delayed to re-render and thus uncaught. In this PR we call the `render` function in advance to get its returned value so we make sure at least the function itself is correct. Then errors in the returns react component will be caught by the error boundary I believe.

### DIFF
--- a/src/canvas/hooks/useRenderWidgetListener.tsx
+++ b/src/canvas/hooks/useRenderWidgetListener.tsx
@@ -127,11 +127,15 @@ export default function useRenderWidgetListener(
         return;
       }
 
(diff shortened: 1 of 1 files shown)


Charlie-XIAO:
This PR is tested on Windows and function correctly. We've got no reviewers for now so merging.

PR #65, opened 2024-06-12
FIX manager layout broken on MacOS

Charlie-XIAO:
See https://github.com/CSCI-SHU-410-SE-Project/Deskulpt/pull/62#pullrequestreview-2112259557 and the following image. There is a margin around the body of the window, and also causes the main contents of the window to be mispositioned. This PR should fix both.

de65c6eb6405dedc3f6bbb94659d3f3

Pinging @ROMEEZHOU for a review to confirm that the problem is fixed on MacOS.

### DIFF
--- a/src/manager/App.tsx
+++ b/src/manager/App.tsx
@@ -33,42 +33,49 @@ export default function App({ initialSettings }: ManagerAppProps) {
   useUpdateSettingListener(setManagerWidgetStates);
 
   return (
-    <Theme appearance={themeAppearance} accentColor="indigo" grayColor="slate">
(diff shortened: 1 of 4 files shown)


ROMEEZHOU (approved):
Works well on Mac! : )

Charlie-XIAO:
@ROMEEZHOU Would you kindly share a screen shot here as well?

ROMEEZHOU:
sure
Screenshot 2024-06-13 at 21 45 54

Charlie-XIAO:
Thanks, looks good and merging

PR #68, opened 2024-06-17
FIX widget errors caught by error boundary persists after fixing and rerendering

Charlie-XIAO:
## Reproducing the error on `main`

Try an arbitrary widget. Make an error in it that will be caught by the error boundary (for instance, using an undeclared variable in the react component which will give a `ReferenceError`). You will then see the error display on the canvas. Now fix that error and click re-render in the manager, the error does not go away.

## Explanation of the fix

In each error boundary fallback we try to reset the error boundary. I'm not sure about the mechanism behind the issue, but the obvious workaround is to try triggering a re-render. The problem with directly calling `resetErrorBoundary` in the fall back is that it will go into an infinite re-rendering loop until the error in the widget gets fixed, because it will try resetting, then detects the error, and falls back, and try resetting again. To avoid this we use a flag variable for which the details are in the comments of the code.

Now with this PR try the reproduction steps again. Re-rendering after fixing the error should automatically update the canvas display and render the correct, working widget.


Charlie-XIAO:
I'm going to merge without approval. This is also unlikely to have OS-specific issues.

PR #182, opened 2024-08-18
MAINT update `time` crate to resolve macOS build failure

Charlie-XIAO:
Action failure: https://github.com/CSCI-SHU-410-SE-Project/Deskulpt/actions/runs/10442469585/job/28915320334

See https://github.com/rust-lang/rust/issues/125319, the recommended fix is to update the `time` crate to work around the known breakage in type inference.


Charlie-XIAO:
Merging as CI passes.

PR #215, opened 2024-08-28
RFC split bundler code into multiple submodules

Charlie-XIAO:
This is a subtask of #66 to avoid crowding a single PR. Essentially, this:

- replaces `bundler.rs` with `bundler/mod.rs` which is equivalent
- moves custom AST transforms in `bundler/mod.rs` to `bundler/transforms.rs`
- moves everything else except the `pub(crate) bundle` function in `bundler/mod.rs` to `bundler/common.rs`
- decompose three subroutine functions that will be reused in #66:
  - `bundle_into_raw_module` bundles entry point into a raw AST module without transforms
  - `apply_basic_transforms` applies TypeScript and JSX transforms
  - `emit_module_to_buf` emits a module to a writable buffer

This is just a pure refactoring of code and functionalities should not change. In particular, none of the unit tests for the bundler is modified so if we are passing in CI we are likely good to go :)


Charlie-XIAO:
CI is passing, let's see how this fits with #66.

PR #216, opened 2024-08-28
ENH support absolute path imports

Charlie-XIAO:
This is part of #66, where we used the absolute path to point to the external dependencies bundle.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -489,6 +489,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "chrono-tz"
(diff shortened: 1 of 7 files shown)


Charlie-XIAO:
On second thoughts, if there is ever a workaround not to use the absolute file in #66, we should close this one. It is not reasonable to use absolute path imports in widgets; after all the imports cannot go beyond the root thus can always be substituted with relative imports...

Charlie-XIAO:
I'm going to close this one. There is a way to accomplish #66 without having to support absolute paths.

PR #217, opened 2024-08-29
TST remove unnecessary exact checks from bundler tests

Charlie-XIAO:
This PR mainly makes the following changes:

- Removed `external_deps` and converted part of it into `default_deps`. The part specifically for external dependencies will be covered in #66 where the new behavior is implemented.
- Combine `replace_apis` into `default_deps` because `@deskulpt-test/apis` is also a default dependency and the complexity clearly allows to use one test for both of them.
- Skip most error messages in the error chain that does not include useful information, such as `load_transformed failed`.
- Use regex for error messages that include paths, URLs, etc. This is mainly to reduce complexity.
- Improve output of `assert_err_eq` a bit to provide more information of what we actually got

### DIFF
--- a/src-tauri/tests/fixtures/bundler/import_url/input/index.jsx
+++ b/src-tauri/tests/fixtures/bundler/import_url/input/index.jsx
@@ -1,1 +1,1 @@
-import osName from "https://cdn.jsdelivr.net/npm/os-name@6.0.0/+esm";
+import osName from "https://foo.js";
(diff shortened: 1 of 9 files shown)


src-tauri/tests/fixtures/bundler/import_url/input/index.jsx
```diff
@@ -1 +1 @@
-import osName from "https://cdn.jsdelivr.net/npm/os-name@6.0.0/+esm";
+import osName from "https://foo.js";
```
Charlie-XIAO, line 1: We have changed to use regex, so that the `+` character in the original URL causes problem (well we can just escape that but let's make things simpler).

PR #224, opened 2024-08-30
MAINT apply TypeScript and React transforms directly in path loader

Charlie-XIAO:
Reference: https://github.com/swc-project/swc/blob/f584ef76d75e86da15d0725ac94be35a88a1c946/crates/swc_bundler/tests/common/mod.rs#L90

The main reason is that in newer versions of SWC the bundler seems unable to parse TypeScript syntax directly so it would be better to strip them off early. There's no harm in doing this - it even saves the effort of doing `apply_basic_transforms` everytime after getting the bundled raw module.


Charlie-XIAO:
Note that we have also removed the verbatim syntax in the typescript transform. All tests are passing so we may wait and see if there are any problems in the actual use.

PR #326, opened 2024-12-11
FIX Config file creation may fail if its parent directory (app_config_dir) doesn't exist.

Xinyu-Li-123:
As is described in the title.


Charlie-XIAO (approved):
LGTM, I've never thought that this directly may not exist.

PR #327, opened 2024-12-12
FIX bug in system info api where CPU info is not aggregated correctly

Xinyu-Li-123:
This PR fix the system api. The `sysinfo` crate suggests [in their doc](https://docs.rs/sysinfo/latest/sysinfo/#usage) to 
> keep the same instance of [System](https://docs.rs/sysinfo/latest/sysinfo/struct.System.html) around instead of recreating it multiple times 

For now, one `System` instance is created everytime the `get_system_info` tauri command is invoked. This makes the fetched information very unstable. For instance, the average CPU usage would fluctuate very rapidly. 

This PR create a static instance of `System` to solve this issue. Lazy evaluation via `once_cell` is used to create this instance. This is because Rust requires static variable's value to be determined at compile time, and lazy evaluation can defer the actual construction to runtime.


Charlie-XIAO (commented):
Nice observation! There are just a few concerns:

- I wonder if with async commands #218 we need to change the lock mechanism?
- Is it possible (or is it necessary) to avoid the unwrap and propagate the error?

Xinyu-Li-123:
You are correct. I have switched to `async_std::sync::Mutex` instead, and remove the unwrap.

src-tauri/src/apis/sys/apis.rs
```diff
@@ -1,10 +1,19 @@
 //! This module implements the commands for `fs` in `@deskulpt-test/apis`.
 
 use crate::commands::CommandOut;
+// use lazy_static::lazy_static;
```
Charlie-XIAO, line 4: Seems to be redundant line

Charlie-XIAO (commented):
LGTM, thanks very much! Just some final nitpicks.

In the future we may want to further refactor sysinfo APIs to obtain them at finer granularity.

src-tauri/src/apis/sys/apis.rs
Xinyu-Li-123: Thanks, it's fixed now.

Charlie-XIAO (approved):

PR #338, opened 2024-12-24
MAINT switch to nightly rustfmt

Charlie-XIAO:
`rustfmt` has too many useful options unstable (thus unavailable on the stable channel). Unlike #337 that switched the whole toolchain to nightly, this PR only uses nightly toolchain for `rustfmt` (which does automatically install the entire nightly toolchain when developing locally, but other components are not affected, i.e., still uses stable).

- `group_imports = "StdExternalCrate"`: Sorts the imports by std-external-crate (hard restriction is always good regarding formatting so we need to care less).
- `imports_granularity = "Module"`: Groups imports per-module so that there is fewer git diff/conflicts on the imports than current.
- `wrap_comments`: I believe Xinyu has complained about rustfmt not being able to wrap comments automatically. It is not rustfmt being unable to do so - it is that there are objections that blocked it from being stabilized (and same happens for many other rustfmt features).

The line width related options are removed. Let's just stick to Rust's default. The formatter will take care of everything.

Note that I only specified `nightly` but not a specific nightly, because unstable features of rust will work no matter which version it is as long as it is nightly. And since we are not relying on the nightly toolchain for other components I imagine this is not a problem.


Charlie-XIAO:
@Xinyu-Li-123 this one should be easy to review.

Charlie-XIAO:
I'm going to bypass the merging rules - there is not likely going to be objections (I think Xinyu never cares about the actual formatting as long as the formatter can handle everything 🤣). This is however blocking some further maintenance effort so I may merge now as is.

PR #353, opened 2024-12-26
ENH do not rely on import map for dynamic import resolution

### LINKED ISSUE #44: RFC/ENH importmap not supported on some machine
I was testing Deskulpt on a Ubuntu20 virtual machine, and I find that importmap is unsupported by the version of webview I am using. I think we should directly replace `@deskult-test/xxx` with relative path, instead of relying on importmap to resolve named import. We are already doing this with `@deskulpt-test/apis`, which is replaced with a blob url, we might as well do it for all default dependencies.

I think we could create a hash map that maps named imports of default dependencies to the corresponding relative path. For example, 
```json
{
    "@deskulpt-test/react": "/assets/src/.scripts/react.js",
    "@deskulpt-test/raw-apis": "/assets/src/.scripts/raw-apis.js",
    "@deskulpt-test/apis": "/assets/src/.scripts/apis.js",
}
```
Based on whether we are building or developing, we can load different maps (what `importmapPlugin` does currently).

Charlie-XIAO:
Fixes #44. The mechanism is approximately as follows:

- The scripts are made entrypoints so that they will be preserved in the build output even if they are not imported in the codebase (this is the same as before).
- The scripts are built at `[name].js` rather than `assets/[name].js` so as to keep consistency between dev and build (this is the same as before).
- Entry signatures are preserved so that the exports of the scripts are as-is (to be properly imported at runtime). This is the same as before, but note that we are no longer using `preserveModules: true` (it seems not necessary now, but I cannot remember exactly why we needed it previously in #37).
- Thanks to `import.meta.url` ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta#url)), we can get the URL base so as to resolve the paths of the imports. Note that since we are importing from a blob, directly using e.g. `/.scripts/jsx-runtime.js` will lead to `TypeError: Failed to resolve module specifier "/.scripts/jsx-runtime.js". Invalid relative url or base scheme isn't hierarchical.`, but the import will work if we have `${baseUrl}/.scripts/jsx-runtime.js`.
- For the widget APIs blob, the URL to the raw APIs is resolved with `baseUrl` from the template.
- For the widget blob, `baseUrl` is passed to the backend bundler and the corresponding imports are replaced directly with the resolved import paths.


Charlie-XIAO:
@Xinyu-Li-123 Would you like to take a look?

Issue #360, opened 2025-01-12
Ideas for dealing with external dependencies


Charlie-XIAO:

Charlie-XIAO:
[Rolldown](https://github.com/rolldown/rolldown) is a bundler with similar functionalities as rollup, but written in Rust which means we can directly use it into our backend. It is not yet stabilized thus not published to crates.io, but we can depend on it via git.

PR #362, opened 2025-01-13
REFACTOR use rolldown to replace SWC for bundler

### LINKED ISSUE #16: BUG Deskulpt will crash when widget code contains duplicate `export default`
If a widget source code contain duplicate `export default` statements, Deskulpt will crash. This is not a bug of Deskulpt, but a bug of SWC, as is discussed in #11. https://github.com/swc-project/swc/issues/8824 related to this bug has been submitted to the SWC repo.

Charlie-XIAO:
Towards #360. Closes #16 as a side effect.

- [x] Replace SWC code with rolldown.
- [x] Rewrite a simplified version of rolldown alias plugin.

Since rolldown has a lot of functionalities out of the box (compared with the SWC bundler which is not well-developed), the new version of bundler only takes ~150 lines of code.

## Future Work

- Initialize a bundler per widget on per widget collection refresh instead of on per rendering. Currently the bundling only depends on widget configuration (change on collection refresh), base URL (static), and APIs blob URL (may change on collection refresh). This would requires some refactoring on the current widget state management.
- Enable watch mode. I haven't experimented, but the rolldown watcher takes `Arc<Mutex<Bundler>>` which means that we can share a bundler instance with the watcher, so that manual re-render can still work as well. Based on the watch mode we can implement live reloading which should greatly enhance widget development experience.
- Work out widget packaging/distribution and how to use external dependencies without having `node_modules`. See a potential design in https://github.com/CSCI-SHU-410-SE-Project/Deskulpt/issues/360#issuecomment-2585681930.


crates/deskulpt-core/src/path.rs
```diff
@@ -23,6 +23,7 @@ pub trait PathExt<R: Runtime>: Manager<R> {
     fn init_widgets_dir(&self) -> Result<()> {
         let widgets_dir = WIDGETS_DIR.get_or_init(|| {
             let resource_dir = self.path().resource_dir().unwrap();
+            let resource_dir = dunce::simplified(&resource_dir);
```
Charlie-XIAO, line 26: `rolldown` does not seem to work with the verbatim prefix `\\?\` on Windows. `dunce::simplified` will remove that prefix whenever possible.

PR #424, opened 2025-02-20
feat: make settings accessible to widget components

### LINKED ISSUE #328: How can a widget hook a custom handler to container's event?
I'm writing a widget that looks like [xeyes](https://linux.die.net/man/1/xeyes), a pair of eyes that tracks mouse movement. To achieve this, I need both the mouse position and widget container position. I'm stuck at how to hook a custom event handler to the `onDrag` event of widget container. This would be trivial if I have access to the widget container, because I could just pass in a callback function to the `onDrag` prop of react draggable. 

For now, I get the container position with `useRef`, and refresh eye parameters periodically, but that feels like the wrong way to do this.

For reference, here is a gif of the widget

And this is my current implementation of xeyes widget.

```typescript
/**
 * Xeyes, but in Deskulpt
 */

import React from "@deskulpt-test/react";

const Eye = ({ mouseX, mouseY, containerX, containerY, left, top }) => {
  const [pupilPos, setPupilPos] = React.useState({ left: 0, top: 0 });

  React.useEffect(() => {
    const dx = mouseX - containerX;
    const dy = mouseY - containerY;
    const angle = Math.atan2(dy, dx);
    const radius = 20; // Distance the pupil can move within the eye
    const left = radius * Math.cos(angle);
    const top = radius * Math.sin(angle);
    setPupilPos({ left, top });
  }, [mouseX, mouseY, containerX, containerY]);

  return (
    <div style={{
      width: 50,
      height: 50,
      borderRadius: '50%',
      backgroundColor: 'white',
      border: '2px solid black',
      position: 'absolute',
      left: left,
      top: top,
      display: 'flex',
      alignItems: 'center',
      justifyContent: 'center',
    }}>
      <div style={{
        width: 20,
        height: 20,
        borderRadius: '50%',
        backgroundColor: 'black',
        position: 'relative',
        left: pupilPos.left,
        top: pupilPos.top,
      }} />
    </div>
  );
};

const XEyes = () => {
  const containerRef = React.useRef<HTMLDivElement>(null);
  const [mousePos, setMousePos] = React.useState({ x: 0, y: 0 });
  const [containerPos, setContainerPos] = React.useState({ x: 0, y: 0 });

  React.useEffect(() => {
    const handleMouseMove = (event) => {
      setMousePos({ x: event.clientX, y: event.clientY });
    };
    window.addEventListener('mousemove', handleMouseMove);
    return () => {
      window.removeEventListener('mousemove', handleMouseMove);
    };
  }, []);

  // Periodically refresh eyeX and eyeY using the container's position
  React.useEffect(() => {
    const interval = setInterval(() => {
      if (containerRef.current) {
        const rect = containerRef.current.getBoundingClientRect();
        setContainerPos({ x: rect.left, y: rect.top });
        console.log(`Container position: ${rect.left}, ${rect.top}`);
      }
    }, 2000); // Update every 100ms

    return () => clearInterval(interval);
  }, [containerRef]);

  return (
    <div ref={containerRef} style={{ width: 300, height: 200 }}>
      <Eye 
        mouseX={mousePos.x} mouseY={mousePos.y} 
        containerX={containerPos.x} containerY={containerPos.y} 
        left={0} top={0}
        />
      <Eye 
        mouseX={mousePos.x} mouseY={mousePos.y} 
        containerX={containerPos.x + 50} containerY={containerPos.y} 
        left={50} top={0}
        />
    </div>
  );
};

export default {
    render: () => <XEyes />,
    width: "105px",
    height: "55px",
}

```

Charlie-XIAO:
Closes #328.

This PR makes the settings available to widgets, currently including `x`, `y` (widget position) and `opacity`.

A few considerations:

- Do we need to expose `opacity` as well for now? I think there is no harm doing so though.
- This will trigger rerendering of widgets as they move. Though moving is considered rare case, probably we want to allow widgets to opt-out of rerendering if they do not actually depend on certain properties. One simple solution would be to let widgets export a list of dependencies e.g. `["x", "y"]` and we use `useMemo` with a comparison function based on a dependencies list. A (potentially better) solution is to analyze which attributes they depend on directly from source code (e.g. `() => {}` means no dependency, `({ x }) => {}` means it only depends on `x`, `(props) => {}` means it depends on all properties). This is harded implementation-wise but would free users from specifying another field. But anyways I think this is out of the scope of the current PR. If it is worth it, we can open a separate issue for this.


Charlie-XIAO:
```tsx
import { useState, useEffect } from "@deskulpt-test/react";

const Eye = ({ left, top, angle }) => {
  const [pupilPos, setPupilPos] = useState({ left: 0, top: 0 });

  useEffect(() => {
    const radius = 20; // Distance the pupil can move within the eye
    const pupilLeft = radius * Math.cos(angle);
    const pupilTop = radius * Math.sin(angle);
    setPupilPos({ left: pupilLeft, top: pupilTop });
  }, [angle]);

  return (
    <div
      style={{
        width: 50,
        height: 50,
        borderRadius: "50%",
        backgroundColor: "white",
        border: "2px solid black",
        position: "absolute",
        left: left,
        top: top,
        display: "flex",
        alignItems: "center",
        justifyContent: "center",
      }}
    >
      <div
        style={{
          width: 20,
          height: 20,
          borderRadius: "50%",
          backgroundColor: "black",
          position: "relative",
          left: pupilPos.left,
          top: pupilPos.top,
        }}
      />
    </div>
  );
};

const XEyes = ({ x, y }) => {
  const [mousePos, setMousePos] = useState({ x: 0, y: 0 });

  useEffect(() => {
    const handleMouseMove = (event) => {
      setMousePos({ x: event.clientX, y: event.clientY });
    };
    window.addEventListener("mousemove", handleMouseMove);
    return () => {
      window.removeEventListener("mousemove", handleMouseMove);
    };
  }, []);

  const leftEyeX = x + 25;
  const rightEyeX = x + 75;
  const centerX = (leftEyeX + rightEyeX) / 2; // Midpoint between both eyes
  const centerY = y + 25;
  const dx = mousePos.x - centerX;
  const dy = mousePos.y - centerY;
  const angle = Math.atan2(dy, dx); // Single angle for both eyes

  return (
    <div style={{ width: 300, height: 200 }}>
      <Eye left={0} top={0} angle={angle} />
      <Eye left={50} top={0} angle={angle} />
    </div>
  );
};

export default {
  Component: XEyes,
  width: "105px",
  height: "55px",
};
```

Xinyu-Li-123 (approved):
LGTM

PR #449, opened 2025-05-07
Open url in default browser instead of in the webview window

### LINKED ISSUE #258: BUG Hyperlink not function correctly
If I click on a hyperlink in a widget, instead of opening it in default system browser, Deskulpt will open the webpage inside the Canvas webview window, occluding both desktop and the widget with the webpage. I believe this is not an intended feature.

Xinyu-Li-123:
Fixes #258.

This PR addresses issue #258, where hyperlinks were incorrectly opened in the webview window that contains the widgets instead of the system’s default browser. I added the [tauri opener plugin](https://v2.tauri.app/plugin/opener/), and added permission to open url in default browser for both canvas and manager window.

For some reason, if I only add the permission `opener:allow-default-urls`, which should allow opening urls with default browser app, the hyperlink won't be opened at all. So I have to use the permission `opener:default`.

To test this feature, you could write a custom widget that contains an element like this 

```html
<a href="https://example.com" target="_blank">click here</a>
```

The `target="_blank"` is necessary for this link to be opened by default browser.


crates/deskulpt/Cargo.toml
```diff
@@ -14,7 +14,9 @@ deskulpt-core                  = { workspace = true }
 tauri                          = { workspace = true, features = ["macos-private-api"] }
 tauri-plugin-clipboard-manager = { workspace = true }
 tauri-plugin-global-shortcut   = { workspace = true }
-
+tauri-plugin-opener            = { workspace = true }
+# required by tauri-plugin-opener
+serde_json = { workspace = true }
```
Charlie-XIAO, line 19: This is strange spacing here. You shall move `serde_json` up and keep the space between `dependencies` and `dev-dependencies`. Note that there are tools for removing unused dependencies so most cases we don't need that comment, but feel free to keep it if you feel necessary.

crates/deskulpt/capabilities/canvas.json
```diff
@@ -4,5 +4,5 @@
   "description": "Capabilities of the canvas window.",
   "local": true,
   "windows": ["canvas"],
-  "permissions": ["core:event:default"]
+  "permissions": ["core:event:default", "opener:default"]
```
Charlie-XIAO, line 7:
```suggestion
  "permissions": [
    "core:event:default",
    "opener:allow-open-url",
    "opener:allow-default-urls"
  ]
```

> For some reason, if I only add the permission `opener:allow-default-urls`, which should allow opening urls with default browser app, the hyperlink won't be opened at all. So I have to use the permission `opener:default`.

This is because you are missing the `opener:allow-open-url` permission. You may check its [initialization script](https://github.com/tauri-apps/plugins-workspace/blob/a6b854032d0b10f0f17c4ffa6bdf4a05429e05fb/plugins/opener/guest-js/init.ts), where under the hood it calls `plugin:opener|open_url`.

crates/deskulpt/capabilities/manager.json
```diff
@@ -4,5 +4,9 @@
   "description": "Capabilities of the manager window.",
   "local": true,
   "windows": ["manager"],
-  "permissions": ["core:event:default", "clipboard-manager:allow-write-text"]
+  "permissions": [
+    "core:event:default",
+    "clipboard-manager:allow-write-text",
+    "opener:default"
```
Charlie-XIAO, line 10:
```suggestion
    "opener:allow-open-url",
    "opener:allow-default-urls"
```

Same as above.

src/package.json
```diff
@@ -13,6 +13,7 @@
     "@radix-ui/themes": "^3.2.0",
     "@tauri-apps/api": "^2.5.0",
     "@tauri-apps/plugin-clipboard-manager": "^2.2.0",
+    "@tauri-apps/plugin-opener": "^2.2.6",
```
Charlie-XIAO, line 16: This is needed only for API and at least in the current version of the PR I don't see it used anywhere.

Charlie-XIAO (commented):
Thanks for the PR! I can confirm it work. A few considerations here:

> The `target="_blank"` is necessary for this link to be opened by default browser.

I tried removing `target="_blank"` and I wasn't able to get back to my canvas again - is this expected? If so I think this is fatal because users can forget to add `target="_blank"` and that should not break our application. In [this script](https://github.com/tauri-apps/plugins-workspace/blob/a6b854032d0b10f0f17c4ffa6bdf4a05429e05fb/plugins/opener/guest-js/init.ts) it explicitly ignores the case `!(a.target === "_blank" || evt.ctrlKey || evt.shiftKey)`. We could consider vendoring this script and change the behavior if we want. WDYT?

src/package.json
Xinyu-Li-123: Fixed.

crates/deskulpt/capabilities/manager.json
Xinyu-Li-123: Fixed.

crates/deskulpt/capabilities/canvas.json
Xinyu-Li-123: Thanks for the information. Fixed.

crates/deskulpt/Cargo.toml
Xinyu-Li-123: By spacing, do you mean the spacing of `serde_json` or the spacing between `dependencies` and `build-dependencies`? I have fixed the latter. The former is to pass taplo's check. If I format it like this

```toml
tauri-plugin-opener            = { workspace = true }
# required by tauri-plugin-opener
serde_json                         = { workspace = true }
```

the format check won't pass.

Charlie-XIAO: I was talking something about this:

```toml
[dependencies]
deskulpt-core                  = { workspace = true }
serde_json                     = { workspace = true }  # required by tauri-plugin-opener
tauri                          = { workspace = true, features = ["macos-private-api"] }
tauri-plugin-clipboard-manager = { workspace = true }
tauri-plugin-global-shortcut   = { workspace = true }
tauri-plugin-opener            = { workspace = true }
```

Xinyu-Li-123: I have tried this earlier and it will also fail the taplo format check. Removing the comment at the line of `serde_json` will pass the check though. As a workaround, I have to add it to the end.

Xinyu-Li-123:
Also, I encountered a strange problem when adding the `tauri-plugin-opener` dependency. I originally only added `tauri-plugin-opener = { workspace = true }` in `crates/deskulpt/Cargo.toml` under `[dependencies]`, similar to other tauri plugins. However, that caused a compile error about `serde_json` being an undeclared crate. The exact error message is this

```
error[E0433]: failed to resolve: could not find `serde_json` in the list of imported crates
  --> crates/deskulpt/src/lib.rs:63:14
   |
63 |         .run(generate_context!())
   |              ^^^^^^^^^^^^^^^^^^^ could not find `serde_json` in the list of imported crates
   |
   = note: this error originates in the macro `generate_context` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0433]: failed to resolve: could not find `serde_json` in the list of imported crates
  --> crates/deskulpt/src/lib.rs:63:14
   |
63 |         .run(generate_context!())
   |              ^^^^^^^^^^^^^^^^^^^ could not find `serde_json` in the list of imported crates
   |
   = note: this error originates in the macro `generate_context` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this struct
   |
7  + use std::iter::Map;
   |
```

I'm not sure exactly why this happens, but explicitly adding the `serde_json` dependency in the toml file fixes it.

Charlie-XIAO:
That's normal. If some macro expands to use `serde_json::xxx` then if we don't have it explicitly in `dependencies`, even if that crate already depends on `serde_json` it will not work.

Unless the crate is carefully designed like `pub use serde_json::*` and in the macro use `crate::serde_json::xxx`, downstream crates will need to explicitly introduce `serde_json` in `dependencies`.

Xinyu-Li-123:
> That's normal. If some macro expands to use `serde_json::xxx` then if we don't have it explicitly in `dependencies`, even if that crate already depends on `serde_json` it will not work.

Are you saying that `tauri-plugin-opener` uses a macro that expands to `serde_json::xxx`? If that is the case, then why does the error occur at the macro `tauri::generate_context`?

Charlie-XIAO:
You can try expanding that macro recursively (rust-analyzer can do that), then you will see when registering opener plugin in `tauri::ipc::RuntimeAuthority`, its `global_scope_schema` uses `::serde_json::xxx`.

Charlie-XIAO:
Note that all other plugins we rely on does not has `global_scope_schema: None` in the expanded macro, except for opener that has `global_scope_schema: Some(...)`. So the problem is with `tauri` itself, and this error will likely occur for any plugin that has a `global_scope_schema`.

Charlie-XIAO:
And with these being said, regarding the `Cargo.toml` formatting, I'm fine with (1) your current style, but remember to remove trailing whitespace in that comment line (not sure why taplo did not detect that) (2) just remove the comment and move `serde_json` up.

Charlie-XIAO (approved):
LGTM, thanks!

Do you have any opinion on https://github.com/CSCI-SHU-410-SE-Project/Deskulpt/pull/449#pullrequestreview-2830557418 by the way?

Xinyu-Li-123:
> Do you have any opinion on [#449 (review)](https://github.com/CSCI-SHU-410-SE-Project/Deskulpt/pull/449#pullrequestreview-2830557418) by the way?

Yes, I'm still thinking about if there is a way to do this without maintaining our own fork of the opener plugin. That might add more burden to future maintenance since we need to keep up to date with any changes made to the plugin.

Charlie-XIAO:
https://github.com/tauri-apps/plugins-workspace/blob/a6b854032d0b10f0f17c4ffa6bdf4a05429e05fb/plugins/opener/guest-js/init.ts

Agreed. But IMO we don't need to fork, we just need to copy the logic in the script above. It actually has very little to do with the plugin itself, only the `invoke("plugin:opener|open_url", { url })`, which can be replaced by `openUrl(url)` if we import `@tauri-plugin/opener`. Other parts of the logic are just very standard frontend stuff.

Xinyu-Li-123:
I think an easier approach would be to listen on all click event, and add `target: _blank` to the `<a>` element. We could use the same logic as the plugin to obtain the `<a>` element from the event. Maybe adding a hook to both manager and canvas like this

```typescript
export function useEnforceOpenNewTab() {
    useEffect(() => {
        const handleClick = (event: MouseEvent) => {
            const link = event
                .composedPath()
                .find((el) => el instanceof Node && el.nodeName.toUpperCase() === 'A') as
                | HTMLAnchorElement
                | undefined

            if (link && link instanceof HTMLAnchorElement && link.href) {
                link.setAttribute('target', '_blank');
            }
        };

        document.addEventListener('click', handleClick);
        return () => document.removeEventListener('click', handleClick);
    }, []);
}
```

Xinyu-Li-123:
> But IMO we don't need to fork, we just need to copy the logic in the script above.

Oh, wait. By this do you mean we could write a custom `init.ts`, and use the backend of the opener plugin?

Charlie-XIAO:
Indeed layering a custom hook could be better. If I'm understanding correctly, you rely on the fact that document event listeners runs before window event listeners in the buffle phase?

Charlie-XIAO:
> > But IMO we don't need to fork, we just need to copy the logic in the script above.
> 
> Oh, wait. By this do you mean we could write a custom hook based on `init.ts`, and use the backend of the opener plugin?

Yes.

Xinyu-Li-123:
> Indeed layering a custom hook could be slightly better regarding maintainability. If I'm understanding correctly, you rely on the fact that document event listeners runs before window event listeners in the buffle phase?

I don't see how it relates to my proposed code.

Charlie-XIAO:
Not sure if I read correctly, but your code is adding the `_blank` target per click? If this is fired after the listener of the plugin, then it will not see the `_blank` target and still open in the webview.

Xinyu-Li-123:
Yes, it does depends on this.

Charlie-XIAO:
But well it seems that you misunderstood my previous idea. And I think this hook has a similar amount of vendored code as directly copying `init.ts`, I currently still lean towards the latter. And I don't think we actually need to make it a hook? It does not have to be in React after all. My idea is that:

- Copy this `init.ts` somewhere into our codebase under `src/` (maybe `utils` or whatever better name), with the `_blank` handling tweaked.
- Import in both `canvas/main.tsx` and `manager/main.tsx`.

Alternatively we can do make it an initialization script and insert into the window on the Rust side, see `crates/deskulpt-core/src/window`.

Xinyu-Li-123:
Just to be clear, are you leaning toward the solution that registers a handler for `click` event that immitates the logic of `init.ts`? If so, is it possible for link with `_blank` to be opened twice, given that we haven't unregister the handler registered by opener plugin?

Charlie-XIAO:
`open_js_links_on_click` can be turned off. `tauri_plugin_opener::Builder::new().open_js_links_on_click(false).build()`

Xinyu-Li-123:
Yeah, that makes sense. I will try to implement this.

Xinyu-Li-123:
I have implemented an event handler that imitates the init script of the opener plugin. The main change is removal of this check 

```typescript
// or not supposed to be open in a new tab
|| !(
  a.target === '_blank'
  // or ctrl key pressed
  || evt.ctrlKey
  // or shift key pressed
  || evt.shiftKey
)
```

Could you take a look @Charlie-XIAO ?

Charlie-XIAO:
LGTM! Just did some minor cleanup to remove `target="_blank"` usage within our codebase since they are no longer necessary. I'm enabling auto-merge 👍

PR #451, opened 2025-05-12
doc: refactor with vitepress and new deployment workflow

Charlie-XIAO:
Overrides #60.

This PR uses [VitePress](https://vitepress.dev/) to refactor the Deskulpt main website. Though VitePress is Vue-based, we would hardly ever need to write Vue scripts. VitePress is more modern and easier to configure than Docusaurus from my own experience. To serve the docs locally:

```bash
pnpm install
pnpm docs:dev # press 'q' to stop the server
```

Screenshot of the homepage (light and dark mode):

image

**About the new docs deployment workflow:** We ultimately want the website to be at the root of `.github.io` instead of under `.github.io/Deskulpt`, because ideally out final website would be `https://deskulpt.github.io/`. If we continue with this monorepo approach, we would have to deploy the pages from the `.github.io` repo. So the overall flow is this: (1) CI of each PR only builds the site (2) CI of main builds the site and pushes the artifacts to `.github.io` repo, in this case, https://github.com/CSCI-SHU-410-SE-Project/CSCI-SHU-410-SE-Project.github.io/, (3) the `.github.io` repo will deploy to GitHub pages. I have set up the SSH public key under the `.github.io` repo with its private key stored in the secrets of this repo. This is also my first time playing with such a setup, so I do not guarantee it to work at first attempt. But anyways, we can only know whether it is working after this PR gets merged.


Xinyu-Li-123 (approved):
LGTM.

PR #457, opened 2025-05-13
Use `document_dir` instead of `resource_dir` for the base directory of widget directory

### LINKED ISSUE #454: AppImage of Deskulpt panic on Ubuntu 22.04
When building from the latest version of Deskulpt (b5deefbad10e36456aabeb10405951792f490976), I find that the AppImage crashes on Ubuntu 22.04. The error is

```text
thread 'main' panicked at /home/eiger/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tauri-2.2.0/src/app.rs:1149:11:
Failed to setup app: error encountered during setup hook: Read-only file system (os error 30)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Aborted (core dumped)
```

After investigating, I find that AppImage mounts its embedded SquashFS filesystem to a temporary read-only location. This makes `WIDGET_DIR` read-only as well. As an example, `WIDGET_DIR` is at `/tmp/.mount_DeskulcTPUT6/usr/lib/Deskulpt/widgets` in my case. This happens because Deskulpt uses `tauri::path::PathResolver::resource_dir` as the base directory of widget directory.

Installing Deskulpt from a deb file works normally, since the `WIDGET_DIR` is `/usr/lib/Deskulpt/widgets` in that case. And my guess is that this works normally on MacOS and Windows as well.

Xinyu-Li-123:
Fixes #454.

This PR addresses issue #454. Additionally, it conditionally uses `resource_dir` when running in `dev` profile (e.g. `pnpm tauri dev`), allowing us to separate debugging widgets from those used in production.


Charlie-XIAO (approved):
Thanks for the PR! Mostly LGTM, I'm okay with this approach, or at least let's try for some time and see if this is the most comfortable development workflow. Just some minor suggestions to simplify the code.

Charlie-XIAO:
BTW, never mind the `docs-build` pending check in the CI.

Charlie-XIAO (approved):
LGTM, merging 👍

PR #544, opened 2025-10-05
ci: completely remove release cache and disable release build in pr workflow

Charlie-XIAO:
- Do not cache release at all. This saves ~3GB of cache from the 10GB cache limit. Also this means that when our dependencies and crates increase in the future, we get 3x cache size increase rather than 6x.
- In normal PR workflow, disable ci-build completely. Add seperate frontend and backend builds in ci-verify instead. ci-build can still be triggered in PR via `[ci-build]` in commit message, but note that it would have no cache so pretty slow. ci-conclude will not depend on ci-build so that normal PRs can still pass quickly.

### DIFF
--- a/.github/actions/setup-rust/action.yaml
+++ b/.github/actions/setup-rust/action.yaml
@@ -64,11 +64,12 @@ runs:
         targets: ${{ inputs.targets }}
 
     - name: Setup cache
-      uses: swatinem/rust-cache@v2
+      uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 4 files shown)


Charlie-XIAO:
It seems `[ci-build]` flag is not working, because `github.event.head_commit.message` in pull requests are the head of the merge branch rather than head of the pr branch. To be fixed in another PR because currently we don't have cache and CI takes too long.

PR #580, opened 2025-10-15
fix: cannot resolve the generated modules with pnpm tauri dev

Charlie-XIAO:
#528 broke the generated modules. In particular, when you do `pnpm tauri dev`, you will see that widgets have errors like `Failed to fetch dynamically imported module: blob:http://localhost:1420/9a490e78-3362-4378-b2f8-b28552e723d4`. This does not happen when you do `pnpm tauri build ...`.

If you inspect the console of the canvas window, you will see errors like `http://localhost:1420/gen/react.js:1   Failed to load resource: the server responded with a status of 404 (Not Found)`. This is because with the vite development server, it directly looks at the source and those files are at `src/gen/...` rather than `gen/...`, while vite build maps them to `gen/...` inside `dist/`. This PR moves `src/gen/...` files to `gen/...` so that the resolution of vite dev and vite build are consistent.


Charlie-XIAO:
@vivienhenz24 Can you check if `pnpm tauri dev` on this PR works for you as well? Thanks!

PR #585, opened 2025-10-15
feat: add hot reload on src code change

vivienhenz24:
implement hot reload feature on code change in widget src folder

### DIFF
--- a/packages/deskulpt-manager/src/components/Settings/HotReload.tsx
+++ b/packages/deskulpt-manager/src/components/Settings/HotReload.tsx
@@ -1,0 +1,40 @@
+import { Select } from "@radix-ui/themes";
+import { deskulptSettings } from "@deskulpt/bindings";
+import { useSettingsStore } from "../../hooks";
+import { useCallback } from "react";
+import { logger } from "@deskulpt/utils";
+
+const options = [
+  { value: "on", label: "On" },
+  { value: "off", label: "Off" },
+] as const;
+
+const HotReload = () => {
+  const hotReloadEnabled = useSettingsStore((state) => state.hotReloadEnabled);
+  const triggerStyle = { width: "100px" };
+
+  const onValueChange = useCallback((value: "on" | "off") => {
+    deskulptSettings.commands
+      .update({ hotReloadEnabled: value === "on" })
+      .catch(logger.error);
+  }, []);
+
+  return (
+    <Select.Root
+      size="1"
+      value={hotReloadEnabled ? "on" : "off"}
+      onValueChange={onValueChange}
+    >
+      <Select.Trigger style={triggerStyle} />
+      <Select.Content>
+        {options.map((option) => (
+          <Select.Item key={option.value} value={option.value}>
+            {option.label}
+          </Select.Item>
+        ))}
+      </Select.Content>
+    </Select.Root>
+  );
+};
+
+export default HotReload;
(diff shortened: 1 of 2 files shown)


vivienhenz24:
still got some things to clean up but opened it for homework's sake

crates/deskulpt-core/src/commands/bundle_widgets.rs
```diff
@@ -77,6 +75,6 @@ pub async fn bundle_widgets<R: Runtime>(
         .into_iter()
         .collect::<HashMap<_, _>>();
 
-    RenderWidgetsEvent(reports).emit_to(&app_handle, DeskulptWindow::Canvas)?;
+    app_handle.emit_on_render_ready(RenderWidgetsEvent(reports))?;
```
vivienhenz24: Switching to app_handle.emit_on_render_ready(RenderWidgetsEvent(reports))? lets the command respect the initial-render guard that buffers widget updates until the canvas window has finished wiring its render-widgets listener. When hot reload now calls the shared bundle_widgets command directly from the watcher thread, it can fire before the canvas has invoked setRenderReady, which would drop the payload if we emitted immediately. By routing through emit_on_render_ready, we reuse the existing state that either delivers the event right away or stores it until the canvas is ready, preventing missed updates during startup.

vivienhenz24:
> Another thing, which we can discuss more in the issue:
> 
> For this PR, I'm okay with just recursively watching everything. However ideally I think we should use the rolldown watcher for each widget and we only watch top level subdirectories. In particular, watch `widgets_dir`, on folder creation/deletion/rename/etc., trigger rescan of widgets. For each widget, then use a `rolldown::Watcher`. Since rolldown is primarily designed for web, I expect we need to make some upstream PRs to rolldown to enhance its watcher capabilities (especially regarding lifetime management). I'm not sure if this is doable or if this is the correct idea, so we should definitely have some discussions in #562!
> 
> For the roadmap of that issue, I would expect something like:
> 
> * This PR
> * Add a settings to turn on/off hot reload
> * The rolldown watcher thing, or improvements to our watcher (depending on which path we decide to go)
> * Any other things you come up with related to this feature

Okay sounds good! Let's not merge this in then, I'm going to spend tomorrow writing up the issue properly and then we can think on how to build on the changes I already made/or take a different direction if we see that's more fitting. Does that sound ok?

vivienhenz24:
issues to resolve
---

There are two blocking issues in the hot reload PR:
- Only deletion works on Windows, other events don't seem to work, the handler doesn't even capture them correctly; file system watching is a very subtle task and this is impossible to fix very quickly
- Deletion is also using refresh; this is incorrect because refresh tries to render after reload, and for deletion the render will fail, though users cannot notice that
So we probably will not be able to merge that PR before presentation, unfortunately. 😢


Oh and another problem: install/uninstall/upgrade widgets will touch the widgets directory and manually do a reload/refresh. This will cause double reload/refresh (or even triple, because the fs notifier could have emitted muliple events during the download process). We cannot just remove that manual reload/refresh either because IMO this hot reload should be an optional feature.


----

Charlie-XIAO:
And there is another suggestion here: if I'm not mistaken notify's watchers can dynamically watch/unwatch paths after creation, so I think we can achieve our goal with only one watcher (rather than maintaining multiple watcher instances which could be more complex and higher overhead).

PR #588, opened 2025-10-16
refactor: replace `with_widget_catalog` with returning lock guard

Charlie-XIAO:
The `with_` style taking a closure looks ugly and does not fit well into control flow. Returning lock guard is better, and this is what we do for `Settings` so keeping the consistency is also nice.


Charlie-XIAO:
This is blocking so I'm gonna bypass rules and merge. It doesn't fundamentally change any functionalitiy so should be fine.

PR #599, opened 2025-10-20
fix: proper timing of initial rescan of widgets

Charlie-XIAO:
Towards #559.

I'm actually not sure if this is the best and most extensible design, but this is already the best I can think of after a bunch of trials and errors. Any suggestion of better ideas are welcome 🙏

So the idea of this PR is, there is a central registry `SetupTasks` per frontend window. For each event listener, we "register" its event name in this registry, and on setup complete we "complete" this task. Every time we "complete" a task, the registry will compare the number of registered tasks versus the number of completed tasks, and once they match, it calls the `complete_setup` command.

The `complete_setup` command distinguishes the two frontend windows: canvas and manager. The `complete_setup` method provided by `SetupStateExt` returns true if all windows have completed setup after the call. This has a few indications (which may or may not be desired):

- In dev mode, our app runs in react strict mode so the `useEffect` hooks are called twice. This means that the `rescan_widgets` command would be triggered twice (which IMO is fine, and after all this behavior is dev only).
- If the user reload either the canvas or the manager window, the `rescan_widgets` command will be triggered again. Though this wasn't expected by me when I design this solution, now I actually think it is a pretty good behavior, since even in production build, webview does have that refresh button, and this can ensure that if user refreshes the canvas webview, the widgets will still be automatically rendered (but well, this point is probably arguable).


vivienhenz24:
what about registering/unregistering from inside the effect so that only hooks that actually mount contribute to the expected-task count?

Charlie-XIAO:
Quote Vivien:

> What I suggested wouldn't work and I can't think of a better approach than the PR.

I'm merging then.

Issue #561, opened 2025-10-09
feat: add logging support


vivienhenz24:
Deskulpt needs structured logging and crash reporting to aid debugging and monitoring. The backend should use Rust's tracing with contextual fields (widget_id, plugin, context), log to rotating files in PERSIST_DIR/logs/ (max 10MB per file, keep last 10, auto-cleanup), and capture panics. The frontend should have global error handlers (exceptions, promise rejections, console errors) and React error boundaries with source maps. Integrate Sentry for crash reporting with PII filtering and opt-in telemetry. Implement log collection commands (list files, read entries, get stats, clear) and a log viewer UI in Manager with search/filter, color-coded levels, and CSV export. Logs could be JSON? (one per line) with timestamp, level, message, and structured fields. Ensure cross-platform support (Windows, macOS, Linux), minimal overhead, privacy-first (no PII), and user control with a telemetry toggle in Settings and clear disclosure.

Charlie-XIAO:
For easier review process, let's do this in separate steps/PRs (not necessarily ordered):

- Bare minimum: Set up logging (with tracing) in the backend, replace the `eprintln!`s and `println!`s, and add a log tab in the frontend which is only capable of viewing all logs.
- Add the `log` command for redirecting frontend console messages and errors to the backend logging implementation.
- UI/UX: Support opening the logs directory.
- UI/UX: Support cleaning up logs.
- UI/UX: Support filtering.
- UI/UX: Support changing log level (this should be a setting). Possibly related: [tracing_subscriber::reload::Handle::modify](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/reload/struct.Handle.html#method.modify).
- Anything else you want to add that would fit into this issue.
- Telemetry (open a separate issue for this).

WDYT @vivienhenz24? Feel free to dump your thoughts here (I believe you have done more investigation than me for this feature). If you have nothing to add, please still leave a comment here, and ping me whenever you have PRs ready for review! Thanks!

vivienhenz24:
@Charlie-XIAO generally agree with your conception of how it should look like. And i think your breakdown into incremental PRs makes sense
The only thing I have to say is that maybe instead of separate “Logs” or “Telemetry” sections, we could group everything under a single diagnostics panel in settings with tabs for logs, crashes, and telemetry. I think it would be a cleaner UX.

[...]

Issue #614, opened 2025-10-26
refactor: split deskulpt-core into multiple internal tauri plugins


Charlie-XIAO:
## Background

We currently have huge `deskulpt-core` plugin. It was mainly a temporary workaround as a proof of concept that the Tauri plugins approach can work. Now that we are deciding to add a bunch of new features and it's time for a refactoring to make code structure clearer. Currently there are a few problems:

- The states extensions are directly applied on apps and app handles. This requires us to add a bunch of prerequisite traits on these extension traits. This not only makes the declarations long and ugly, but we can also easily forget to clean up when a prerequisite trait is no longer necessary. Moreover, we have to properly namespace the methods to avoid collision. The recommended approach by Tauri is to have a struct that carries an app handle directly, and the extension trait just needs to provide a method that returns this struct. All methods are then implemented on this struct rather than on the extension trait. This solves all our problems. For an example, see how tauri-plugin-dialog or tauri-plugin-global-shortcut are implemented.
- Many of our commands and events are also currently forced to be namespaced. For instance, `bundle_widgets` and `rescan_widgets` could have been `bundle` and `rescan` under a `deskulpt-widgets` plugin.
- Many initialization work that does not require ordering can be completed in plugin setup rather than in app setup. Only those with strict ordering requirements or that depends on other plugins to have already been initialized should be left in the app setup.
- If we found that several functionalities that should be split into separate plugins are too tightly coupled that makes it hard to do so, most likely our design is suboptimal (engineering-wise, not performance-wise) and it is a good chance to decouple them.

### `deskulpt-settings`

This is likely the lowest level plugin. Most other functionalities would need to depend on settings. This plugin might provide an `update` command and an `update` event.

There are many important refactoring that might need to be made. In particular, currently on shortcuts and canvas imode change we call some non-low-level functions. This could potentially be solved with a subscriber pattern, i.e., expose a method for registering to certain updates. This way we remove the dependency of `deskulpt-settings` on higher-level functionalities (decoupled).

### `deskulpt-widgets`

WIP

PR #642, opened 2025-11-02
feat(logging): initialize tracing backend with rotation and panic capture

vivienhenz24:
Towards #561.

@Charlie-XIAO WDYT?

could you run pnpm tauri dev, make sure it compiles on your machine and verify persist_dir/logs/app.log appears?

### DIFF
--- a/crates/deskulpt-core/Cargo.toml
+++ b/crates/deskulpt-core/Cargo.toml
@@ -27,7 +27,7 @@ tokio                        = { workspace = true }
 tracing                      = { workspace = true }
 tracing-appender             = { workspace = true }
 tracing-panic                = { workspace = true }
-tracing-subscriber           = { workspace = true, features = ["env-filter", "time"] }
+tracing-subscriber           = { workspace = true, features = ["time"] }
 
 tauri = { workspace = true, features = [
   "specta",
--- a/crates/deskulpt-core/src/logging.rs
+++ b/crates/deskulpt-core/src/logging.rs
@@ -1,9 +1,0 @@
-use anyhow::Result;
-use tauri::{App, Runtime};
-
-use crate::states::LoggingStateExt;
-
-/// Legacy entry point that delegates to [`LoggingStateExt::manage_logging`].
-pub fn init<R: Runtime>(app: &mut App<R>) -> Result<()> {
-    app.manage_logging()
-}
--- a/crates/deskulpt-core/src/path.rs
+++ b/crates/deskulpt-core/src/path.rs
@@ -13,6 +13,7 @@ static WIDGETS_DIR: OnceCell<Arc<PathBuf>> = OnceCell::new();
 
 /// Thread-safe lazily-initialized static for the persistence directory.
 static PERSIST_DIR: OnceCell<Arc<PathBuf>> = OnceCell::new();
+
 /// Thread-safe lazily-initialized static for the logs directory.
 static LOGS_DIR: OnceCell<Arc<PathBuf>> = OnceCell::new();
 
@@ -98,13 +99,13 @@ pub trait PathExt<R: Runtime>: Manager<R> {
     }
 
     /// Initialize the logs directory.
+    ///
+    /// This will create the logs directory if it does not exist. It must be
+    /// called before calling the [`logs_dir`](PathExt::logs_dir) method.
     fn init_logs_dir(&self) -> Result<()> {
         let logs_dir = LOGS_DIR.get_or_init(|| {
-            let base_dir = self
-                .path()
-                .app_log_dir()
-                .unwrap_or_else(|_| self.path().app_local_data_dir().unwrap().join("logs"));
-            Arc::new(base_dir)
+            let logs_dir = self.path().app_log_dir().unwrap();
+            Arc::new(logs_dir)
         });
 
         if !logs_dir.exists() {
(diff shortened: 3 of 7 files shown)


Charlie-XIAO (commented):
Thanks for the PR @vivienhenz24, the log files do appear. However there are a few things to note before I do more detailed review:

- As I mentioned in one of your previous PRs, we don't put heavy logics in the `deskulpt` main crate. They should currently stay in `deskulpt-core` (and as #614 proceeds they should likely go into a separate `deskulpt-logging` crate but this is not something to care about in this PR).

- Let's not make things overly complicated:
  - `tracing-subscriber` already has a solid `.json()` formatter. IMO reimplementing formatting/visitation adds surface area and future maintenance with little gain. Do you have a good reason for this?
  - I'm not sure why you manually implementing a complex renaming/linking/pruning strategy and the 30ms maintenance loop while `tracing-appender` already provides a sane implementation. What problem are you solving and again, do you have a good reason for reinventing the wheels?

I have not yet looked into more details but I think the above needs to be resolved in the first place for the PR to become cleaner for review :)

vivienhenz24:
yeah you're right I was chasing a flat ts/level/target/message/file/line json shape and a stable app.log tail, but i agree that it's way more machinery than it was worth. Adressing it right now

crates/deskulpt-core/src/path.rs
```diff
@@ -94,6 +96,35 @@ pub trait PathExt<R: Runtime>: Manager<R> {
...
     }
+
+    /// Initialize the logs directory.
+    fn init_logs_dir(&self) -> Result<()> {
+        let logs_dir = LOGS_DIR.get_or_init(|| {
+            let base_dir = self
+                .path()
+                .app_log_dir()
+                .unwrap_or_else(|_| self.path().app_local_data_dir().unwrap().join("logs"));
+            Arc::new(dunce::simplified(&base_dir).to_path_buf())
```
Charlie-XIAO: There is the [app_log_dir](https://docs.rs/tauri/latest/tauri/path/struct.PathResolver.html#method.app_log_dir) API provided by Tauri, and we won't need `dunce::simplified` for it.

crates/deskulpt-core/src/logging.rs
```diff
@@ -0,0 +1,114 @@
...
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::{EnvFilter, Registry, fmt};
+
+use crate::path::PathExt;
+
+const MAX_LOG_FILES: usize = 10;
+
+static INIT: OnceCell<()> = OnceCell::new();
+static LOG_GUARD: OnceCell<WorkerGuard> = OnceCell::new();
+static PANIC_HOOK: OnceCell<()> = OnceCell::new();
```
Charlie-XIAO: - I don't think you would need the `INIT`, because we can guarantee we call `init` only once on Tauri app setup.
- I don't think you would need the `PANIC_HOOK` either, because again we install the hook only once in `init`.
- `LOG_GUARD` indeed needs to be stored for the app's lifetime. A more standard way is to keep it as a state. So what I would recommend in this case is to move the thing to `states/logging.rs` and create a `LoggingState` that wraps the `WorkerGuard`. You can then follow the pattern of other states.

crates/deskulpt-core/src/logging.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    INIT.set(()).ok();
+    Ok(())
+}
+
+fn build_writer(logs_dir: &Path) -> Result<(NonBlocking, WorkerGuard)> {
+    let appender = RollingFileAppender::builder()
+        .rotation(Rotation::DAILY)
+        .max_log_files(MAX_LOG_FILES)
+        .filename_prefix("app.log")
```
Charlie-XIAO: Would `deskulpt.log` make more sense here? And maybe we can use `deskulpt` as prefix and `log` suffix (so that the logs look like `deskulpt.YYYY-MM-DD.log`)?

crates/deskulpt-core/Cargo.toml
```diff
@@ -28,7 +28,11 @@ serde_with                   = { workspace = true }
 serialize-to-javascript      = { workspace = true }
 specta                       = { workspace = true, features = ["derive", "function", "serde_json"] }
 tauri-plugin-global-shortcut = { workspace = true }
+time                         = { workspace = true }
 tokio                        = { workspace = true }
+tracing                      = { workspace = true }
+tracing-appender             = { workspace = true }
+tracing-subscriber           = { workspace = true }
```
Charlie-XIAO:
```suggestion
tracing-subscriber           = { workspace = true, features = ["env-filter", "time"] }
```

This way we won't need to depend on `time`.

Charlie-XIAO (approved):
So I made a few final changes:

- Add some docstrings.
- Cleanup some remaining redundant stuff, like `app_log_dir`.
- Remove `logging::init`: `app_handle.manage_logging` should be able to do its job already. I am assuming this method will be called only once through the app's lifetime (and this indeed is the case).
- Use a deterministic filter rather than an env filter: I believe we will use settings to control log level in the future rather than environment variables.
- Put tracing initialization all into one single function - it turns out to be just ~30 lines, IMO no need to split (variable namings are not ambiguous, structure is still pretty clear, and we need to write fewer types).

@vivienhenz24 You can take a final look. If everything looks good to you, you can merge yourself 🎉 If there are any issues please feel free to ping me again. I will also leave some further comments in #561 to dump some of my further thoughts around the logging feature.

PR #694, opened 2025-11-21
refactor: use `shadow(false)` for canvas instead of manually calling `objc2`

### LINKED ISSUE #676: refactor: macos replace manual `msg_send!` with higher level `objc2_app_kit` bindings
We currently have this:

```rust
        #[cfg(target_os = "macos")]
        {
            use objc2::msg_send;
            use objc2::runtime::{AnyObject, Bool};

            // Disable the window shadow on macOS; there will be shadows left on
            // movement for transparent and undecorated windows that we are using;
            // it seems that disabling shadows does not have significant visual impacts
            unsafe {
                let ns_window = canvas.ns_window()? as *mut AnyObject;
                let () = msg_send![ns_window, setHasShadow:Bool::NO];
            }
        }
```

I imagine this can be replace with: https://docs.rs/objc2-app-kit/latest/objc2_app_kit/struct.NSWindow.html#method.setHasShadow. Tauri gives us `*mut c_void`: https://docs.rs/tauri/latest/x86_64-apple-darwin/tauri/webview/struct.WebviewWindow.html#method.ns_window, I imagine this can be unsafely casted to `NSWindow`, then we can safely call `setHashShadow`.

**I could be wrong though!**

ROMEEZHOU:
Closes #676 . Remove the original logic, directly call shadow() on builder to avoid shadow around widget on macos. 
shadow() function reference: https://docs.rs/tauri/latest/tauri/webview/struct.WebviewWindowBuilder.html#method.shadow


Charlie-XIAO:
Just in case I forget, please change to `Closes #676` in the top comment so that merging this PR can automatically close that issue.

Charlie-XIAO (approved):
I can confirm `.shadow(false)` doesn't break anything on my Windows 11 machine and Ubuntu 24.04 Desktop VM. Thanks @ROMEEZHOU!

PR #698, opened 2025-11-23
feat: show welcome widget on startup

### LINKED ISSUE #686: feat: add a welcome/starter widget containing starter guides
Currently users will be extremely confused of what to do when they launch Deskulpt. In fact, they might not even notice where Deskulpt is (because initially there are no widgets, and the manager is hidden by the system tray). We can add a welcome/starter widgets including some very brief guidance of how to get started. For instance, Rainmeter has this: https://docs.rainmeter.net/manual/getting-started/using-rainmeter/.

vivienhenz24:
Added persistent has_seen_starter_tutorial flag to settings and patch API, plus binding updates. Gated welcome widget in the widgets manager: drops it from the catalog when the flag is true, sets the flag on first load, and logged the gating paths for visibility.

Closes #686

### DIFF
--- a/crates/deskulpt-settings/src/settings.rs
+++ b/crates/deskulpt-settings/src/settings.rs
@@ -177,7 +177,7 @@ pub struct Settings {
     #[serde_as(deserialize_as = "MapSkipError<_, _>")]
     pub widgets: BTreeMap<String, WidgetSettings>,
     /// Whether the starter widgets have been added.
-    #[serde(skip_serializing)]
+    #[specta(skip)]
     pub starter_widgets_added: bool,
 }
 
--- a/docs/src/public/settings-schema.json
+++ b/docs/src/public/settings-schema.json
@@ -33,7 +33,7 @@
     "starterWidgetsAdded": {
       "description": "Whether the starter widgets have been added.",
       "type": "boolean",
-      "writeOnly": true
+      "default": false
     }
   },
   "$defs": {


Charlie-XIAO (commented):
@vivienhenz24 Thanks for the PR, but before I review, please first self-review and clean things up. There are some clearly AI-written rubbish that are completely irrelevant (such as copying over `deskulpt.svg` and `deskulpt-dark.svg`, removing legacy `index.js`, which makes no sense). Also as I've said multiple times, never put heavy logic directly in the `deskulpt` main crate.

crates/deskulpt-widgets/starter/welcome/deskulpt.widget.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "Welcome",
+  "entry": "index.tsx",
+  "description": "Tutorial widget for Deskulpt",
+  "ignore": true
```
Charlie-XIAO: - Let's use `index.jsx` for now. TypeScript support is incomplete currently, and that if you put `index.tsx`, you commonly would need a `tsconfig.json` as well, so I would rather not bother with that currently.
- The description, maybe it should be "Welcmome widget for Deskulpt."? (Note the trailing period as well, descriptions are preferred to be full sentences).
- Should not set `ignore` here. Actually, I'm planning on removing that option and replace with a load/unload setting (rather than a manifest option).

crates/deskulpt-settings/src/settings.rs
```diff
@@ -204,6 +207,9 @@ pub struct SettingsPatch {
     /// patch will be applied to it.
     #[specta(optional, type = BTreeMap<String, Option<WidgetSettingsPatch>>)]
     pub widgets: Option<BTreeMap<String, Option<WidgetSettingsPatch>>>,
+    /// If not `None`, update [`Settings::has_seen_starter_tutorial`].
+    #[specta(optional, type = bool)]
+    pub has_seen_starter_tutorial: Option<bool>,
```
Charlie-XIAO: Similarly, I would mark this with `#[serde(skip)]` (and then the specta attribute will likely not be needed).

crates/deskulpt-widgets/src/catalog.rs
```diff
@@ -158,7 +158,7 @@ impl WidgetCatalog {
                     patches.insert(id.clone(), None);
                 },
                 itertools::EitherOrBoth::Right((id, _)) => {
-                    patches.insert(id.clone(), Some(Default::default()));
+                    patches.insert(id.clone(), Some(WidgetSettingsPatch::default()));
```
Charlie-XIAO: Unrelated change that can be reverted :)

crates/deskulpt-widgets/src/lib.rs
```diff
@@ -21,6 +25,8 @@ deskulpt_common::bindings::build_bindings!();
 pub fn init<R: Runtime>() -> TauriPlugin<R> {
     deskulpt_common::init::init_builder!()
         .setup(|app_handle, _| {
+            app_handle.init_widgets_dir()?;
```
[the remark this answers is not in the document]
vivienhenz24: > Yes that's because the actual state management of the widgets directory happens after the setup of plugins. Personally I would rather not fix like this, but instead make `seed_welcome_widget_if_needed` a method on `WidgetsManager` and call it in the `setup` function in the main `deskulpt` crate, WDYT?

I agree with this. just opened up an issue about the macos slow loading, going to debug right now and open a separate PR (hopefully if i find the root cause)

crates/deskulpt-widgets/Cargo.toml
```diff
@@ -13,6 +13,7 @@ version    = { workspace = true }
 [dependencies]
 anyhow            = { workspace = true }
 bitflags          = { workspace = true }
+copy_dir          = "0.1"
```
Charlie-XIAO: Please specify the version in the workspace root `Cargo.toml` and reference using `{ workspace = true }` here. Also as a binary (not library), we commonly prefer specifier up to patch, i.e., "0.1.3" in this case.

Charlie-XIAO (commented):
Great, this PR is so much clearer now, thanks for your effort @vivienhenz24! Some final comments before I do more detailed review and check the UI:

Charlie-XIAO:
The starter widget currently looks like the following. We will definitely improve in the future, when we have better documentation and the widget gallery set up. I think it could be a book-like thing where one can switch pages.

image

image

PR #726, opened 2025-12-09
docs: add widget installation guide and malware warning

vivienhenz24:


Charlie-XIAO:
We will be refactoring docs largely (and actually move to another repo) so I will close this one for now, but I will keep this in mind and make sure there is something similar added to the new docs. Thanks anyways!

PR #769, opened 2026-01-22
refactor: introduce react-compiler and remove useCallback/memo

### LINKED ISSUE #767: refactor: consider react compiler
https://react.dev/learn/react-compiler

Charlie-XIAO:
Closes #767.


Charlie-XIAO:
Note that not all `useCallback`s can be removed. For instance, those returned by custom hooks, or those used as other hooks' deps.

PR #825, opened 2026-04-25
chore: migrate to tsdown for building libraries

Charlie-XIAO:
Closes #775.


Charlie-XIAO:
Minification of the re-export pattern is worse than rollup + terser, but I believe that would eventually be improved. See oxc's issue-13047 (I intentionally did not link).

